PR #69, opened 2016-03-23
Update strings.xml

XuToTo:
Make some zh-translation words more friendly :)


iSoron:
Thank you for the pull request. Could you please also modify the strings at [POEditor](https://poeditor.com/join/project/8DWX5pfjS0)? The files are automatically generated by POEditor every time a new version of the app is released, so your modification might get overwritten if they are not also uploaded there.

XuToTo:
I have already modified the strings at POEditor.   :smile:

Issue #74, opened 2016-04-03
DAY_OF_WEEK usage


sciamano:
Hi, I was always wondering why in punch card etc, the week starts with SATURDAY (!). Sunday I could may be understand... After digging in the code, I am now worring about line 147 in `DateHelper.java`
`day.set(GregorianCalendar.DAY_OF_WEEK, 0);`
My problem is that according to java docs, the `Calendar.DAY_OF_WEEK` is taking values 1 to 7, not 0. You can check e.g. a code
```
GregorianCalendar day = new GregorianCalendar();
System.out.printf("day = %s, i = %d\n", "SUNDAY", day.SUNDAY);
System.out.printf("day = %s, i = %d\n", "MONDAY", day.MONDAY);
System.out.printf("day = %s, i = %d\n", "SATURDAY", day.SATURDAY);
```

Does it makes more sence to change the line to
`day.set(GregorianCalendar.DAY_OF_WEEK, day.getFirstDayOfWeek());`
can you comment?

...looks like related to #53

iSoron:
Hi sciamano, nice to have someone looking over the source code. Throughout the code, it is assumed that the weekdays are numbered from 0 (Saturday) to 6 (Friday). A couple of places that make this assumption, for example, are [here][as1] and [here][as2]. The decision of assigning zero to Saturday was completely arbitrary, but having them start at zero is usually much more convenient. To keep things consistent, I would like to have every low-level function treat zero as Saturday. The views are then free to pick whatever day is better to start the week.

So, you are right that the line `day.set(GregorianCalendar.DAY_OF_WEEK, 0)` is wrong, since it relies on undocumented behaviour, but I would replace it instead with `day.set(GregorianCalendar.DAY_OF_WEEK, Calendar.SATURDAY)`.

To fix #53 it would be better to change [this method][as3] on HabitHistoryView. Instead of subtracting `todayWeekday` on the last line, we should subtract `todayWeekday + getFirstDayOfWeek` or something like that. We should probably also change [this method][as4]. Instead of starting at zero, `i` should start at `getFirstDayOfWeek`.

Please feel free to ask me any questions if something is still not clear.

[as1]: https://github.com/iSoron/uhabits/blob/dev/app/src/main/java/org/isoron/uhabits/models/RepetitionList.java#L128
[as2]: https://github.com/iSoron/uhabits/blob/dev/app/src/main/java/org/isoron/uhabits/helpers/DateHelper.java#L188-L215
[...]

PR #77, opened 2016-04-05
First day of week

### LINKED ISSUE #53: Allow user to pick Monday or Sunday as the first day of the week
The individual habit statistics would look more familiar to Europeans, if the "saturday" and "sunday" would be at the bottom and "monday" at the top (see [here](http://i.imgur.com/pwupLvL.png) for clarification). A setting, to specify the beginning of the week, could easily correct that.

The calendar view is also somehow irritating, because the months name isn't above the column, where the month actually starts, as shown in the picture above.

sciamano:
Fixes #53 and #74 
the first day of the week will be selected according to the current locale
PS: the history view calculations are very cryptic...

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,5 @@
 # Loop Habit Tracker
-<a href="https://circleci.com/gh/iSoron/uhabits/tree/dev">
-  <img src="https://img.shields.io/circleci/project/iSoron/uhabits/dev.svg">
-</a>
-<a href="https://codecov.io/github/iSoron/uhabits?branch=dev">
-  <img src="https://img.shields.io/codecov/c/github/iSoron/uhabits.svg" alt="Coverage via Codecov" />
-</a>
+## +++ This is a fork +++
 
 Loop is a simple Android app that helps you create and maintain good habits,
 allowing you to achieve your long-term goals. Detailed graphs and statistics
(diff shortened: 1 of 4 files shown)


iSoron:
Thank for the pull request. I will review it soon.

iSoron:
Hi Denis, I had a look at your code. It's a good start, and it works well for most cases, but there are a couple of issues we need to fix before merging this. I am testing it with a locale that has Saturday as the first day of the week.

1. When the current day is Saturday, the leftmost column of the HabitHistoryView becomes blank. There should never be any blank columns.

2. When the history editor is open, clicking on any Saturday toggles a repetition for the Saturday of the following week. For example, clicking March 19 toggles March 26.

3. The previous issue is true with check marks added through the main screen / widgets too. A check mark on March 19 will show up in HabitHistoryView as a checkmark on March 26.

Please let me know if you can't reproduce these issues, or if you get stuck fixing them.

sciamano:
Hi Alinson,  you are right, I can reproduce all the issues. There my second try, that will fix all of them

PS: please exclude the commit with README.md

iSoron:
Nice, it seems to be working fine now. I have made some minor code changes and merged it into `dev`. See 24c02605. Thanks again for the pull request. This is a feature that many users asked for. Please, feel free to add youself to [the about screen](https://github.com/iSoron/uhabits/blob/f5f3ec9f88ad90df47b9cd6cb88312ceedc01d6f/app/src/main/res/layout/about.xml).

sciamano:
thanks!

PR #92, opened 2016-04-29
Fix #91 - update gradle version to overcome build error

### LINKED ISSUE #91: Build failed with an exception:
I've just tried to build the latest version of dev:

```
./gradlew build
```

I get the following error message:
```
Plugin is too old, please update to a more recent version, or set ANDROID_DAILY_OVERRIDE environment variable to ....
```

Do you have any idea how to fix this?

luiscruz:
Quick fix by android studio.
In accordance with what is suggested here:
http://stackoverflow.com/questions/29063968/plugin-is-too-old-please-update-to-a-more-recent-version-or-set-android-daily

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.1.0-alpha3'
+        classpath 'com.android.tools.build:gradle:1.5.0'
     }
 }
 


iSoron:
I am bumping the version to 2.1.0 instead. See 0cd4a414389. Thank you for the pull request, anyway.

luiscruz:
Sounds good ;)

Issue #68, opened 2016-03-22
Export CSV files with data for multiple habits


iSoron:
In addition to the individual files already been exported, it would be nice if we could also export some files that contain data for multiple habits. For example:

    checkmarks.csv
    --------------
    date,habit 1,habit 2,habit 3
    2016-01-01,0,0,1
    2016-01-02,0,1,1
    2016-01-03,1,0,1

    scores.csv
    ----------
    date,habit 1,habit 2,habit 3
    2016-01-01,0.30,0.24,0.70
    2016-01-02,0.29,0.24,0.75
    2016-01-03,0.35,0.24,0.80

PKompis:
I am gonna try it.

PKompis:
It is completed i will push it in the next couple of days.

iSoron:
@PKompis Nice, I'm looking forward to it.

JotraN:
Hi, is this issue still open? I'd like to work on it if it is.

iSoron:
@JotraN Yes, it's still open. Please, feel free to work on it.

PR #114, opened 2016-05-25
Export CSV files with data for multiple habits (Issue #68)

PKompis:
Issue 68 solved. From now on when you press "Export as CSV" two more files are created, AllCheckmarks and AllScores.

### DIFF
--- a/app/src/main/java/org/isoron/uhabits/io/HabitsCSVExporter.java
+++ b/app/src/main/java/org/isoron/uhabits/io/HabitsCSVExporter.java
@@ -62,6 +62,18 @@ public class HabitsCSVExporter
         Habit.writeCSV(habits, out);
         out.close();
 
+        //my contribution
+		String filename2 = "AllCheckmarks.csv";
+		new File(exportDirName).mkdirs();
+		FileWriter out2 = new FileWriter(exportDirName + filename2);
+		generateFilenames.add(filename2);
+        CheckmarkList check = new CheckmarkList();
+        check.writeCSVMultipleHabits (habits, out2);
+		out2.close();
+		//until here
+
+
+
         for(Habit h : habits)
         {
             String habitDirName = String.format("%03d %s/", h.position + 1, h.name);
--- a/app/src/main/java/org/isoron/uhabits/io/HabitsCSVExporter.java
+++ b/app/src/main/java/org/isoron/uhabits/io/HabitsCSVExporter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2016 Álinson Santos Xavier <isoron@gmail.com>
+ * Copyright (C) 2016 Γlinson Santos Xavier <isoron@gmail.com>
  *
  * This file is part of Loop Habit Tracker.
  *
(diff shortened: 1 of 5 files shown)


iSoron:
Thank you for the pull request. I will revise it soon.

app/src/main/java/org/isoron/uhabits/io/HabitsCSVExporter.java
```diff
@@ -60,6 +60,20 @@ private void writeHabits() throws IOException
...
+        out = new FileWriter(exportDirName + filename);
+        generateFilenames.add(filename);
+        CheckmarkList check = new CheckmarkList();
+        check.writeCSVMultipleHabits (habits, out);
+
+        filename = "AllScores.csv";
+        new File(exportDirName).mkdirs();
+        out = new FileWriter(exportDirName + filename);
+        generateFilenames.add(filename);
+        ScoreList score = new ScoreList();
```
iSoron: Please, don't instantiate ScoreList and CheckmarkList here. Make the methods `writeCSVMultipleHabits` static instead, so that you can call `Score.writeCSVMultipleHabits`. Also, I would rename the method to simply `writeCSV`. From the arguments, it is clear that it works with multiple habits.

build.gradle
```diff
@@ -3,14 +3,16 @@ buildscript {
...
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.1.0-alpha3'
+        classpath 'com.android.tools.build:gradle:2.1.0'
     }
 }
 
 allprojects {
     repositories {
         jcenter()
+        maven { url "https://oss.sonatype.org/content/repositories/snapshots/" }
```
iSoron: Why are you making changes to the gradle file?

app/src/main/java/org/isoron/uhabits/io/HabitsCSVExporter.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2016 Álinson Santos Xavier <isoron@gmail.com>
+ * Copyright (C) 2016 Γlinson Santos Xavier <isoron@gmail.com>
```
iSoron: This is probably an encoding error. Please revert this line to before and make sure that your text editor is configured to use UTF-8. The same happened in every file.

app/src/main/java/org/isoron/uhabits/models/CheckmarkList.java
```diff
@@ -299,4 +303,62 @@ public void writeCSV(Writer out) throws IOException
...
+        {
+            out.write(String.format(",%s", habit.name));
+        }
+        do
+        {
+            String timestamp = dateFormat.format(new Date(cursor.getLong(0)));
+            out.write(String.format("\n%s", timestamp));
+            for(Habit habit : habits)
+            {
+                out.write(String.format(",%s", getHabitValue(habit.getId().toString(), Long.toString(cursor.getLong(0)))));
```
iSoron: Unless I am mistaken, you are making one query per habit per timestamp. If the user has been using the app for one year, and has 50 habits, this will result in almost 20 thousand queries. You can certainly obtain the same information with one query per habit, instead. In fact, you don't even need to write your own queries. Just call the functions `habit.checkmarks.getAllValues()` and `habit.score.getAllValues()`.

build.gradle
PKompis: i had to change the dependency from classpath 'com.android.tools.build:gradle:2.1.0' to classpath 'com.android.tools.build:gradle:2.0.0'. but i pushed this change to my repo so i had to to change it back to the previously dependency .

app/src/main/java/org/isoron/uhabits/models/CheckmarkList.java
PKompis: Ok. i am gonna try it today

PR #117, opened 2016-05-28
 Export CSV files with data for multiple habits (Issue #68)

PKompis:
First of all I could not use the getAllValues method as you suggested because my method is static as you again suggested and second it would not work for my case.

### DIFF
--- a/app/src/main/java/org/isoron/uhabits/io/HabitsCSVExporter.java
+++ b/app/src/main/java/org/isoron/uhabits/io/HabitsCSVExporter.java
@@ -60,6 +60,18 @@ public class HabitsCSVExporter
         FileWriter out = new FileWriter(exportDirName + filename);
         generateFilenames.add(filename);
         Habit.writeCSV(habits, out);
+
+        filename = "AllCheckmarks.csv";
+        new File(exportDirName).mkdirs();
+        out = new FileWriter(exportDirName + filename);
+        generateFilenames.add(filename);
+        CheckmarkList.writeCSV(habits, out);
+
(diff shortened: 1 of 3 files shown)


iSoron:
Sorry, you are still making too many queries. As I said before, you don't need to write any SQL to implement this feature. You receive a list of habits as parameter. You can call getAllValues on each of them.

Also, there is no need to close a pull request to make changes. If you make any changes to your branch, this pull request will be automatically updated. Closing and creating new pull requests for every change makes it hard for everyone to follow the conversation.

PKompis:
Would you accept as an asnwer an array with the values of all habits? For example in the first cell it will be the array that the getallvalues return. So if the user is using it for 1 year and he has 5 habits we need 365*5 cells. I thought this method because i cannot find a way to write "vertically" to a csv.

iSoron:
> So if the user is using it for 1 year and he has 5 habits we need 365*5 cells

Yes, this is fine. Since each checkmark value is just an `int`, this matrix would still not take a lot of memory. I am mostly concerned about the number of queries, because queries are slow, even using SQLite. If the user has 5 habits, I would like your code to perform at most 5 queries to fetch all the checkmarks. Your code code look like this:

    writeCSV(List<Habit> habits, Writer out):
        create empty list of int[]
        for each(Habit : h) add h.getAllValues() to the list
        print list to CSV

iSoron:
If you are still working on this pull request, please let me know.

PKompis:
I am not working in this pull request. I study for my exams this 
month.

PR #145, opened 2016-07-06
Add 24hr snooze. Fixes #103

### LINKED ISSUE #103: 24hr snooze
For many things one reminder a day is enough for me, but the lowest frequency is 8hrs.

24hr would also have the nice effect of only reminding at the chosen time.

ajbarnes:
These seemed to be the only changes needed to add this functionality. 
A look at the snooze settings confirmed the change took place as expected.


iSoron:
Thank you for the pull request.

PR #151, opened 2016-07-30
relative size of the marks in the frequency view

sciamano:
In this PR I'll like to introduce relative size of the marks based on the maximum frequency.

I have a low frequency habit that I mark once a month. Consequently, in the frequency view, for the whole month I have only one small gray dot at some weekday even though the actual desired frequency was achived (looks sort of depressive :). The reason was that the mark size will become large only if I do it every week and reach the value >=4.
Of course it does not work for low-frequency habits...


iSoron:
Hi Denis, I went ahead and [added your name to the list of developers][1]. I wrote it as "Денис (sciamano)". Please, feel free to change it if you would like. Thank you again for all your contributions.

[1]: https://github.com/iSoron/uhabits/commit/77e0ad007f350621a368a76185d996ac2e958b0e

sciamano:
thanks!

Issue #112, opened 2016-05-24
Use Storage Access Framework when importing/exporting files


iSoron:
When exporting files, we currently generate intents with URI pointing to files, and when importing, we use a custom file picker that is only able to browse a certain folder on the SD card. A better solution, as suggested by @licaon-kter, would be to use the [Storage Access Framework](https://developer.android.com/guide/topics/providers/document-provider.html), instead. In this way, users could import/export directly from/to cloud storage, for example. Our current solution might also [stop working on Android N](https://commonsware.com/blog/2016/03/14/psa-file-scheme-ban-n-developer-preview.html) and will certainly not work properly on the Chromebook, so this is somehow inevitable.

One issue is that the Storage Access Framework is not available below API 19 (KitKat). We currently support all devices starting from API 15 (Ice Cream Sandwich).

regularcoder:
I will make an attempt at this. I will use conditional execution to support devices below API 19, unless you think that is not necessary.

iSoron:
Hi @regularcoder, are you still working on this? This issue is preventing us from targeting API 24 (Nougat).

regularcoder:
Hi, no I have stopped working on this.

iSoron:
Implemented on `dev` branch. The app now targets API 25 (Android 7.1 Nougat)

PR #157, opened 2016-08-04
Request storage permission before showing the import data screen

Jawnnypoo:
Hey there,

My wife was using the app on her 6.0 device, and realized she was not able to use the import screen. Looks like it was missing the storage permission on 6.0+ devices. This PR fixes this

### DIFF
--- a/app/src/main/java/org/isoron/uhabits/activities/settings/SettingsFragment.java
+++ b/app/src/main/java/org/isoron/uhabits/activities/settings/SettingsFragment.java
@@ -19,21 +19,31 @@
 
 package org.isoron.uhabits.activities.settings;
 
-import android.app.backup.*;
-import android.content.*;
-import android.os.*;
-import android.support.v7.preference.*;
+import android.Manifest;
+import android.app.backup.BackupManager;
+import android.content.Intent;
+import android.content.SharedPreferences;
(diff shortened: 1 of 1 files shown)


iSoron:
Hi John, thank you for the pull request. Could you describe more precisely what was the problem before? What happened when she pressed the button to import the data? The app is not supposed to require permission to read external storage. This is intentional, for security and privacy reasons.

Jawnnypoo:
Ahh, I see where the confusion was. After reading the FAQ, I see that you are wanting users to copy their back up file to a specific directory on their device. 

I guess after reading that you could disreguard this PR. But honestly, I feel it would be a better experience to allow the user to navigate through their files to find the back up file to restore from, as you can with this permission.

screenshot_20160804-204305

I think lots of users are not aware of the process of moving files around on Android and maybe do not even have an app installed that would allow them to move files on the filesystem. But at least with this, they would be able to navigate to their downloads folder. I dunno, what do you think? Maybe it is better left as is I guess?

iSoron:
Yes, the current way does not provide the best user experience, but I really don't want to include that permission. At some point in the future, we will switch to the storage access framework #112, so this will not be an issue anymore. Thank you for the pull request, anyway.

PR #165, opened 2016-08-23
Fix #161 reminder disable on archive

### LINKED ISSUE #161: Archived habits with reminders continue creating alerts
I archived an old habit that had a reminder. I continue receiving reminders after archiving. I would expect reminders to be disabled for archived habits.

klanmiko:
Hi there, thought I could help with this small thing.


iSoron:
Thank you for the pull request.

PR #170, opened 2016-09-02
Fix issue#65 by setting launchMode to singleTop

regularcoder:
Set launch mode on MainActivity to singleTop so that a new instance of the activity is not created every time.


iSoron:
Thank you for the simple solution. I have merged it.

PR #173, opened 2016-09-12
Add exporting scores and checkmarks files with multiple habits (#68).

JotraN:
Concerning #68:
> Add writeMultipleHabits to write a scores file and a checkmarks file
> that contains scores and checkmarks of multiple habits.
> Add getTimeframe because it was necessary to get the timeframe between
> all habits so that row data could be populated correctly for habits that started before/after other habits.
> Move writeCheckmarks to below writeScores since it wasn't called until
> after writeScores in writeHabits.
> 
> Add writeScore (ScoreList) and writeCheckmark (CheckmarkList) so that
> individual values could be outputted.
> 
> Add getNewest to get the newest repetition in a list  and to correspond
> with getOldest.
> 
> Add getDaysBetween to DateUtils to quickly get the number of days
> between two timestamps.

I've also noticed some inefficiencies regarding how often computeAll was called (it's called explicitly before writing, then again when querying), but wasn't sure how to handle them.

### DIFF
--- a/app/src/main/java/org/isoron/uhabits/models/ScoreList.java
+++ b/app/src/main/java/org/isoron/uhabits/models/ScoreList.java
@@ -89,6 +89,46 @@ public abstract class ScoreList implements Iterable<Score>
         return s.getValue();
     }
 
+    /**
+     * Returns the list of scores that fall within the given interval.
+     * <p>
+     * There is exactly one score per day in the interval. The endpoints of
+     * the interval are included. The list is ordered by timestamp (decreasing).
+     * That is, the first score corresponds to the newest timestamp, and the
+     * last score corresponds to the oldest timestamp.
+     *
+     * @param fromTimestamp timestamp of the beginning of the interval.
+     * @param toTimestamp   timestamp of the end of the interval.
+     * @return the list of scores within the interval.
+     */
+    @NonNull
+    public abstract List<Score> getByInterval(long fromTimestamp,
+                                                  long toTimestamp);
+
+    /**
+     * Returns the values of the scores that fall inside a certain interval
+     * of time.
+     * <p>
+     * The values are returned in an array containing one integer value for each
+     * day of the interval. The first entry corresponds to the most recent day
+     * in the interval. Each subsequent entry corresponds to one day older than
+     * the previous entry. The boundaries of the time interval are included.
+     *
+     * @param from timestamp for the oldest score
+     * @param to   timestamp for the newest score
+     * @return values for the scores inside the given interval
+     */
+    public final int[] getValues(long from, long to)
+    {
+        List<Score> scores = getByInterval(from, to);
+        int[] values = new int[scores.size()];
+
+        for(int i = 0; i < values.length; i++)
+            values[i] = scores.get(i).getValue();
+
+        return values;
+    }
+
     public List<Score> groupBy(DateUtils.TruncateField field)
     {
         computeAll();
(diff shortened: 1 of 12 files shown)


iSoron:
Thank you for the pull request. I will review it shortly. In the meantime, could you have a look at the failed tests? It seems something went wrong.

JotraN:
Sorry, I forgot to ran the tests.
The issue was I didn't check for when checkmarks did not exist when writing them.
I've fixed the issue and updated and rebased my branch.

iSoron:
Hi @JotraN, I had a look at the code. The code is well written, and you are generating the CSV files correctly. The only problems is that you are performing one query for each habit and for each day. This can get very slow on large databases.

Instead of calling `h.getCheckmarks().getValues(timestamp, timestamp)` for each habit and for each timestamp, you can call `h.getCheckmarks().getAllValues()` for each habit, instead. In this way, you will only require one query per habit.

```
values = []

for(Habit h : habits)
    values.add(h.getCheckmarks().getAllValues())

for(i=0; i<days; i++)
    for(j=0; j<habits.size(); j++)
        write(values[j][i]);
```

For the scores, you could do something similar, but using the method `h.getScores().getAll()`.

JotraN:
@iSoron, thanks for the input!

I'm starting to work on the more efficient solution, but would like your input on my current idea.
I'm thinking of adding a function like this to CheckmarksList and ScoresList, so that I can get the all the values without having to checking to see if they exist in the exporter.
```
int[] getValues(long from, long to)
    // get all the values.
    // prepend values with zeroes for the number of missing days before the habit was created
    // append values with zeroes for the number of missing days after the habit was created
```
However, a getValue function already exists in CheckmarkList, so I'm not sure what I should name this function.

Alternatively, I could just check for missing values in the exporter, but that kinda feels messy.

iSoron:
Actually, this is exactly how the method `checkmarkList.getValues` was supposed to work, but now I see there was a bug in `compute` that prevented it from working. I went ahead and fixed it, since `compute` is one of the most tricky methods in the app. The method `getValues` now returns a padded list of values.

I also fixed `scoreList.compute`, but there is no corresponding `getValues` there, since it was not needed. Feel free to add it. You can have a look at the implementations of `checkmarkList.getValues` to see how is it done. Basically, after calling `compute(from, to)` you can assume that the scores will already be on the database, so you can just query them.

JotraN:
Cool, thanks for the fixes. It made the code a lot easier to implement.
I've added the changes in the most recent commit (after rebasing/merging).

app/src/main/java/org/isoron/uhabits/models/ScoreList.java
```diff
@@ -89,6 +89,24 @@ public final int getValue(long timestamp)
...
     }
 
+    /**
+     * Gets the values of all the scores between the given days.
+     *
+     * @param from the timestamp of the first day
+     * @param to the timestamp of the second day
+     * @return the values of all the scores between the given days
+     */
+    public final int[] getValues(long from, long to)
```
iSoron: This method actually returns the values of all the scores that have already been computed. For example, in the following unit test, I am only interested in the values for 3 days:
```
    @Test
    public void test_getValues()
    {
        toggleRepetitions(0, 20);

        long today = DateUtils.getStartOfToday();
        long day = DateUtils.millisecondsInOneDay;

        long from = today - 4 * day;
        long to = today - 2 * day;

        int[] expected = {
            11883254,
            11479288,
            11053198,
        };

        int[] actual = habit.getScores().getValues(from, to);
        assertThat(actual, equalTo(expected));
    }
```
This unit test fails with this error:
```
java.lang.AssertionError: 
Expected: [<11883254>, <11479288>, <11053198>]
     but: was [<12629351>, <12266245>, <11883254>, <11479288>, <11053198>, <10603773>, <10129735>, <9629735>, <9102352>, <8546087>, <7959357>, <7340494>, <6687738>, <5999234>, <5273023>, <4507040>, <3699107>, <2846927>, <1948077>, <1000000>]
```

JotraN: Woops, sorry about that!

I've fixed the issue and added the test case as well as test cases for the other functions I've added.
I also changed the newest date to use when exporting the csv to the current day, since all the other files use the current day (but I've left getTimeFrame in, in case this needs to be used by another function in the future).

iSoron:
Everything looks good now. Thank you for the patch!

Issue #62, opened 2016-03-22
Increase width of name column according to screen size


sojusnik:
If in the main overview the name of a habit is too long to fit one line, then it should be trimmed with an indicator like "…" at the end of a word to maintain consistency throughout all habits.

Currently, the normal view on my device shows 4 days, the widescreen view entire 11! I think, that it would be more practical to expand the space for the habit's names in the widescreen view, so that only 7 days, instead of the 11, are shown. As a result, you'll get a better overview over habits with longer names.

The space for a habit's name is in my case (OnePlus X with FullHD Resolution) in the widescreen view a little bit smaller than in the normal view. Is this by design?

iSoron:
The name column has a fixed width (roughly), regardless of your device orientation. The remaining space is filled with checkmarks. The larger the screen, the more checkmarks you can see. There is no upper limit. I agree that we could compute this width in a more clever way.

iSoron:
By the way, if anyone has some time, I would love to receive pull requests for small enhancements like this.

PR #174, opened 2016-09-14
Increase width of name column according to screen size

Donaira:
This is regarding the issue: https://github.com/iSoron/uhabits/issues/62

I changed `labelWidth` calculation to take into account screen's width. Now it's either 1/3 screen width OR R.dimen.habitNameWidth depending on which is larger.


iSoron:
Thank you for the patch. It seems to be working fine.

Issue #168, opened 2016-09-01
Change message on home screen widget when habit is deleted


italiatina:
More of a feature request/enhancement

When I add a habit to the home screen as a widget, and then delete that habit, the widget remains on the home screen, but now reads "Error drawing widget."

iSoron:
Thank you for the suggestion. I agree that we should do that, but unfortunately, as far as I know, Android does not allow it. Only the user can add and remove widgets to the home screen. The best we could do is change the message to something more helpful ("Habit deleted / not found").

regularcoder:
I'll take a shot at changing the message to "Habit deleted / not found" on the widget when a habit is deleted.

PR #175, opened 2016-09-17
Fix for issue #168

regularcoder:


iSoron:
Thank you for the patch. I don't think we need the field `habitNotFound`. You are essentially using it to return an extra value whenever `getHabitFromWidgetId` throws an exception. The same could be achieved with a custom exception. You would need to:

1. Make `getHabitFromWidgetId` throw a more specific exception, say `HabitNotFoundException`. You would have to create this exception. It could go into the package `org.isoron.uhabits.models`.
2. Make the function `drawErrorWidget` receive an exception as parameter. If the exception is an instance of `HabitNotFoundException`, then it shows the "not found" message. Otherwise, it shows the generic one.

regularcoder:
Thanks for reviewing. I agree that a custom exception approach is better. Will revise and resubmit.

regularcoder:
Updated to use a custom exception instead of a flag. Please can you review again?

I inherited from RuntimeException for the sake of simplicity.

iSoron:
Looks good. I have merged it. Thank you for the patch.

PR #183, opened 2016-09-29
Fix 'NoSuchMethodError getColor' for API < 23

sirekanyan:


iSoron:
Could you tell me what problem are you trying to solve? Is the app crashing in some particular phone? What phone is it, and what version of Android? Could you send me the stacktrace?

iSoron:
I have fixed this issue in commit 638b3f763c. The solution was not checking `SDK_INT >= M`, but using `ResourcesCompat` instead. Thank you for bringing my attention to this.

Issue #192, opened 2016-10-21
Allow habits to have interval larger than 99 days


akhater:
I have a task that needs to be done every 4 months but the max time I can enter in the flexible schedule is 99 days :)

thanks

aleksejrs:
I also have such a task (although 4 months is the absolute maximum).

akhater:
Hi Aleksejrs 

I'm interested why would you consider 4 months as absolute maximum ? is it from a technical perspective or you do you mean that this is no longer called a habit :)

aleksejrs:
It's not a habit; some companies require specific usage actions to happen at least that often.

iSoron:
Thank you for the suggestion, @akhater. The limit is there because I didn't expect anyone to have habits with such long intervals, but there is no technical reason why these should not be allowed.

Implementation: We just need to increase the maximum length of the text field from 2 to 3 characters. I think 999 is a reasonable upper limit.

iSoron:
Fixed by 10131d5.

akhater:
Thank you

tilt-me:
Seems this issue is back with version 2.0.1: now it is not possibile to set an interval > 99 days:  @iSoron  could you please restore the 3-char limit?  
Thanks in advance, bye!

palashkulsh:
this issue seems to have resurfaced

neube3:
The limit seems to still be at 999.

It's still an arbitrary limit, why not a bigger one? The app is used for remembering about health and medical check-ups can often have more than 999 days between them required (e.g. once every 1825 days/5 years" - checking for Lyme disease markers, cancer screenings, etc.).

While 99999 days (273 years) seems pretty overkill, who's to say the human lifespan won't extend that much in the next 40 years of us using the app?

PR #202, opened 2016-11-01
Fix for #192

jtrollkarl:
- added a separate style specifically for tvFreqDen to allow a habit to
be repeated more than 99 days, thus increasing the cap to 999.


iSoron:
Thanks for the pull request, @jtrollkarl. Could you make both numbers large (not just the denominator)? You can get rid of the old style completely `dialogFormInputSmallNumber`, I think.

Issue #222, opened 2016-12-11
Remove use of static context from HabitsApplication class


AnirudhaAgashe:
Holding a static reference to the context in application can lead to memory leaks.

Also the call back to [onTerminate ](https://developer.android.com/reference/android/app/Application.html#onTerminate())is only received in case of emulated environment and not on actual device.

AnirudhaAgashe:
I was able to replace the static context in most of the classes without any issues by injecting Application and activity context wherever applicable. In case PebbleReceiver used the context that is received in receiveData call back. Also implemented a alternative method of checking isTestMode which doesn't required context.

However in case of ListHabitsController I had to inject activity context. ListHabitsController requires context to carry out ExportCSVTask and ExportDBTask.

Is there any idea how we can inject context for ExportCSVTask and ExportDBTask without having to inject context in ListHabitsController ?

iSoron:
Please, submit a pull request and I will have a look at the modifications you made. We can discuss the code changes there. Are you trying to get rid of the deprecated method `HabitsApplication.getContext()`?

AnirudhaAgashe:
In effect yes. The main cause of leak is the static reference to context which needs to be removed.

PR #224, opened 2016-12-12
Remove use of static context from HabitsApplication class

AnirudhaAgashe:
removed static context from application class. Replaced static method to obtain context with appropriate dependency

### DIFF
--- a/app/src/androidTest/java/org/isoron/uhabits/BaseViewTest.java
+++ b/app/src/androidTest/java/org/isoron/uhabits/BaseViewTest.java
@@ -225,7 +225,7 @@ public class BaseViewTest extends BaseAndroidTest
         throws IOException
     {
         File dir = FileUtils.getSDCardDir("test-screenshots");
-        if (dir == null) dir = FileUtils.getFilesDir("test-screenshots");
+        if (dir == null) dir = FileUtils.getFilesDir(testContext,"test-screenshots");
         if (dir == null) throw new RuntimeException(
             "Could not find suitable dir for screenshots");
 
--- a/app/src/androidTest/java/org/isoron/uhabits/io/ImportTest.java
+++ b/app/src/androidTest/java/org/isoron/uhabits/io/ImportTest.java
@@ -53,7 +53,7 @@ public class ImportTest extends BaseAndroidTest
 
         fixtures.purgeHabits(habitList);
         context = InstrumentationRegistry.getInstrumentation().getContext();
-        baseDir = FileUtils.getFilesDir("Backups");
+        baseDir = FileUtils.getFilesDir(context, "Backups");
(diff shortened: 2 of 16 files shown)


AnirudhaAgashe:
I have closed the issue #222 and generated a pull request with the changes that were needed to remove the static context.

iSoron:
Hi @AnirudhaAgashe, sorry about the long delay. I have been busy recently, but I will review your pull request shortly. Thank you for sending it.

iSoron:
Hi @AnirudhaAgashe, thank you again for the pull request. I agree with the removal of the static method `getContext()` from the class `HabitsApplication`.

I would like to keep the code inside the controllers completely Android-free. That is, to instantiate a controller, we shouldn't need a context. I see you currently use the context when calling `exportCSVFactory.create(context, selected, ...)`. This can be avoided by modifying the constructor of `ExportCSVTask`. If you write it like this:
```
public ExportCSVTask(@Provided @AppContext Context context,
                     @Provided @NonNull HabitList habitList,
                     @NonNull List<Habit> selectedHabits,
                     @NonNull Listener listener)
```
then AutoFactory will automatically find a reference to the context for us. Then, on the controller, you can just call `exportCSVFactory.create(selected, ...)`. The class `ExportDBTask` can also be modified to use AutoFactory, just like ExportCSVTask, so we don't need a reference to the context. Please, let me know if you need any help doing this.

Other than that, please add either `@Nullable` or `@NonNull` to every method parameter.

AnirudhaAgashe:
Hi @iSoron , Sorry for the delay.

I was of the same opinion that Controller should not have context and should be free from any Android dependency. I was actively seeking your input on how should we go about it when I created #222.

This was just meant to give an idea of what needs to be done and not to be actually merged in.
As you said I will remove context from controller and modify the constructor for exportTask and will create a new pull request with all the necessary annotations.

iSoron:
> As you said I will remove context from controller and modify the constructor for exportTask and will create a new pull request with all the necessary annotations.

Great. Just to let you know, you don't need to create a new pull request. After you push new commits to your branch, they will automatically appear here.

AnirudhaAgashe:
I have removed context from controller, modified export csv and db task to use application context via auto factory. Also added annotations. If anything more needs to be done let me know.

iSoron:
Looks great now. Thank you for the patch, @AnirudhaAgashe.

iSoron:
Hi @AnirudhaAgashe, it seems that this patch caused our automated tests to fail. See:

    https://circleci.com/gh/iSoron/uhabits/469

Could you have a look and submit another pull request to fix this?

AnirudhaAgashe:
The testContext needs to be replaced with targetContext in tests. I will do it and submit another pull request to fix it.

PR #231, opened 2017-01-07
Refactoring ColorUtils class

mbartosik:
Little refactoring of ColorUtils class .


iSoron:
Hi @mbartosik, thank you for the pull request, however, generally speaking, I will not accept pull requests from new contributors that simply switch from one programming style to another (in this case, from constants to enums). If you enjoy refactoring code, one thing you could try to work on is improving our test coverage. There are many areas of the app which are currently not being tested at all.

PR #236, opened 2017-01-15
Fixed failing build

AnirudhaAgashe:
Fixed failing build by replacing testContext by targetContext. test context was returning null directory causing the test to fail


iSoron:
Thanks!

PR #258, opened 2017-02-22
Added stackable notifications

recheej:
Hey it's my first commit to an open source project! This seems like a great app to learn from. Let me know what you think of the change. Stackable notifications are available in version 7.1. Since I don't have a 7.1 device, I tested on a virtual device. I'm also new to Android development so any tips/tricks/comments on mistakes I've made would be greatly appreciated.

### DIFF
--- a/app/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsActivity.java
+++ b/app/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsActivity.java
@@ -24,12 +24,8 @@ import android.os.*;
 import org.isoron.uhabits.*;
 import org.isoron.uhabits.activities.*;
 import org.isoron.uhabits.activities.habits.list.model.*;
-import org.isoron.uhabits.models.Habit;
 import org.isoron.uhabits.preferences.*;
 import org.isoron.uhabits.utils.*;
-
-import java.util.ArrayList;
-import java.util.List;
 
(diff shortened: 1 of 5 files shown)


iSoron:
Hi @recheej, thank you for your contribution! I will review it soon.

recheej:
Added a commit I didn't want to this pull request. That check in is now reverted.

recheej:
@iSoron I just made a new commit. I noticed I missed some things. Also, I improved the way I was doing the stackable notifications a bit. Here's the summary:

1. If there's only one 'active' notification, it will show the notification as they were being shown before.
2. If there's more than one active notification, they will be bundled (stacked). 

- On below version 7.0 devices, these bundled notifications show as one 'summary' notification and its content is: 
'[# of reminders] habit reminders'
- If the user expands this summary, they will see each of the individual habit names listed.
- If the user clicks the summary, they will be taken to habit list activity.

Test it out and let me know what you think.

iSoron:
Your patch seems to work well on Nougat. However, on devices older than Nougat, it doesn't seem to be possible to check or snooze the habits from the notification anymore, unless there is only one notification active. Could you change your patch, so that, on devices older than Nougat, everything stays the same as it is today -- that is, each notification is shown by itself?

iSoron:
Closing this pull request due to no response. If you are still working on this, please let me know.

Issue #131, opened 2016-06-19
Fix usage of singular/plural in alert messages


brdvlps:
A minor mistake: After deleting a single habit, I got the message habits (notice the plural) successfully deleted. This confused me and I had to check twice to make sure I haven't made unwanted changes.

Language: German

aleksejrs:
See also issue 132.

iSoron:
Thank you for reporting this issue. I guess there are other messages that would also need to be corrected.

PR #260, opened 2017-02-26
Fix Issue #131

recheej:
Added strings for plural and singular versions of the strings we had. Also added the code to display singular or plural depending on count of the action.

### DIFF
--- a/app/src/main/java/org/isoron/uhabits/commands/ArchiveHabitsCommand.java
+++ b/app/src/main/java/org/isoron/uhabits/commands/ArchiveHabitsCommand.java
@@ -49,13 +49,21 @@ public class ArchiveHabitsCommand extends Command
     @Override
     public Integer getExecuteStringId()
     {
-        return R.string.toast_habit_archived;
+        if(selectedHabits.size() == 1){
+            return R.string.toast_habit_archived;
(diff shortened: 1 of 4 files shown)


iSoron:
Thank you for the pull request, and sorry for the long delay. Although your solution works great for English, it will probably not work very well for some other languages. The recommended way of handling plurals in Android is by using the method `getQuantityText`. Here is what the [official guide has to say](https://developer.android.com/guide/topics/resources/string-resource.html#Plurals).

I suggest the following changes, but if you have a better way of implementing, please feel free to share:

1. Add a method `getQuantity` to each command, which will return the number of objects affected by the command;
2. Change the method `ListHabitScreen.onCommandExecuted`, so that you would call `showMessage(command.getExecuteStringId(), command.getQuantity())`;
3. Create a method `BaseScreen.showMessage(@PluralRes Integer stringId, int quantity)`. This method would be similar to the existing one, but instead of simply calling `snackbar.setText(stringId)`, it would call `snackbar.setText(activity.getResources().getQuantityText(stringId, quantity))` instead.

iSoron:
Closing this pull request due to no response. If you are still working on this, please let me know.

Issue #274, opened 2017-04-12
Reminder days are reset when the reminder time is changed.


artem-p:
I have a habit with a partial-wheek reminder, 5 days a week for example. When I change reminder time, reminder days are reset back to all week.

iSoron:
Thank you for reporting this bug.

iSoron:
Fixed by bb0d43018e.

PR #275, opened 2017-04-12
fix resetting reminder days #274

artem-p:
#274 
I've added a check in onTimeSet method.


iSoron:
I have merged your patch. See bb0d43018e. Thank you for the contribution!

PR #276, opened 2017-04-13
Use of Cyrillic letters to display Bulgarian

mihail-stefanov:
I switched the phonetic notation with Cyrillic letters to denote the Bulgarian language. I hope this is the proper place to offer this amendment.


iSoron:
Thanks!

Issue #99, opened 2016-05-11
More colors for habits


JoshGraham:
(Translating some user feedback from Google Play Store)
...If there were more colors for habit, it'd be great!

iSoron:
I guess we could add a few of the missing [Material Colors](http://www.materialui.co/colors), such as "orange" (we have "deep orange") and brown.

I am not allowing user-defined colors because I need to be able to tweak the colors according to the theme of the application. The colors in night mode, for example, must be much brighter and less saturated than the normal ones. Doing it dynamically doesn't look as good as picking the colors by hand.

YPetremann:
The problem here is that with a lot of task, colors are redondant, for my part, I think we need 2 more grays, one between the two exinsting one and one lighter than the existing,
a brown, a pink, a indigo and a light blue,

PR #286, opened 2017-05-06
More colors for habits

austil:
Issue #99

Hi, I wanted to start contributing to this amazing app I use every day 😄 so here i am !

This "small task" exposed some design issue.
Habit's color are saved as an index of the current color palette and the goal was to add color to this same color palette.

So, I had to choose between three options : 

* Add color at the end of the palette => this makes the color picker ugly
* Add color at their "best position" => this will shift habit's color for current user after update
* Make changes to keep current color's indexes of the database while placing color at their best position

I first tried the third option (4ce9013) but failed (revert 894423e) and went for the second (8a80a66).

What do you think ?

Augustin 🇫🇷

### DIFF
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -20,12 +20,16 @@
         <item>@color/red_700</item>
         <item>@color/deep_orange_700</item>
         <item>@color/orange_700</item>
+        <item>@color/amber_800</item>
         <item>@color/yellow_800</item>
         <item>@color/lime_700</item>
+        <item>@color/light_green_600</item>
         <item>@color/green_700</item>
         <item>@color/teal_600</item>
         <item>@color/cyan_600</item>
         <item>@color/light_blue_600</item>
+        <item>@color/blue_700</item>
+        <item>@color/indigo_700</item>
         <item>@color/deep_purple_600</item>
         <item>@color/purple_600</item>
         <item>@color/pink_600</item>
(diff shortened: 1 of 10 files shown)


iSoron:
Thank you for the pull request, Augustin! I will review it soon. I am fine with adding the new colors in the correct position, even if this means that we will have to make some changes to the database.

app/src/main/res/values/colors.xml
```diff
@@ -28,13 +29,16 @@
         <item>@color/deep_purple_600</item>
         <item>@color/purple_600</item>
         <item>@color/pink_600</item>
+        <item>@color/brown_700</item>
         <item>@color/grey_800</item>
+        <item>@color/grey_600</item>
```
iSoron: Since we are already changing things, I think we should go ahead and add all the colors from the Material Design palette. More specifically, we are still missing Indigo, Blue, Light Blue, Light Green and Amber.

app/src/main/res/values/colors.xml
```diff
@@ -44,13 +48,16 @@
         <item>@color/deep_purple_200</item>
         <item>@color/purple_200</item>
         <item>@color/pink_200</item>
+        <item>@color/brown_200</item>
         <item>@color/grey_100</item>
+        <item>@color/grey_300</item>
```
iSoron: I like the corresponding colors in the dark palette. Good job.

iSoron:
> this will shift habit's color for current user after update

We can solve this by adding a database migration. When users update the app, we can update the numbers on the database in order to keep the colors the same.

In order to do this, first you would need to increment the database version number to 20 (on the file `app/build.gradle`), then create a migration file called 20.sql in `assets/migration` including the SQL update commands. I can do this part, if you are not comfortable with SQL.

austil:
Ok, i think i can do that.

austil:
Finally found 15min to do that so here it is 😃.

Why TeamCity's build are failing ?

iSoron:
Everything looked good, so I went ahead and merged your pull request. Thank you for your contribution! The new colors will appear in version 1.8. If you would like to work on anything else, just let me know.

Issue #283, opened 2017-04-30
Widget showing archived habits


mr-flibble:
Hello, when I create a widget, a list will appear containing all created habits, even those archived.
I think that archived habits should be hidden from this list. It will certainly improve usability.
Thanks.

iSoron:
Thank you for the suggestion, @mr-flibble. I agree that we should not list archived habits in that list.

samagra14:
@iSoron I would love to work on this issue. I am new to open source and therefore targeting small tasks. Can I pick up the issue and work on it? Or is the issue already being handled? If I can work on this issue then I will need your help to get started.

iSoron:
Hi @samagra14, please feel free to work on it.

PR #309, opened 2017-06-19
Don't list archived Habits when creating Widgets.

PrototypeNM1:
Related to issue #283.

Note: I haven't built to verify this works given that 1. the fix is trivial and self explanatory and 2. setting up and building it will take a fair bit longer on my side than yours given my build environment isn't set up. If I'm wrong on either of those two points let me know.


iSoron:
Looks good. Thank you for the pull request! If you would like to work on anything else, just let me know.

PrototypeNM1:
So I just checked this in the current release and it doesn't appear to have solved the issue. Any ideas on why this wouldn't have worked?

PR #312, opened 2017-07-10
Keep Selection while during screen resize or rotation

judeibe:
Eddited androidmanifest to to not reload activity on orientation and screenSize


iSoron:
Thank you for the pull request. I am not sure this is the best way to solve this problem, though. According to the [official documentation](https://developer.android.com/guide/topics/manifest/activity-element.html), using `android:configChanges` should be avoided and used only as a last resort. Although I am sure this solves this issue, I am worried this might break something else.

PR #316, opened 2017-08-01
Viewing

WLun001:
Fixed textsize
Using dp instead of sp for text sizes


iSoron:
Hi @WLun001, what problem are you trying to solve with this patch?

WLun001:
All of the textView in sp would change after changing font size in settings, dp - would not change. 
It is better to use sp for textsize instead of dp.

iSoron:
This part of the code was copied directly from the Android Open Source Project (AOSP) source code. I am not sure why did they use `dp`, instead of `sp` for text, but my guess is that having the text resize automatically would break the picker layout. For this reason, I will not merge these changes. In any case, thank you for the patch, @WLun001.

Issue #319, opened 2017-08-02
"No" action to dismiss a notification if "Make notifications sticky" setting is checked


rsri:
For persisted notifications, if a user wants to dismiss a particular notification (For example, waking up early habit on a Sunday), with the current version, it's impossible. It would be better to have a "No" action for persisted notifications that performs the same action as swiping a non-persisted notification. If the idea is approved, I can take up this tiny feature and finish it. :)

iSoron:
This is related to #307. I think it's okay to add a "no" action to all notifications, even if the notifications are not sticky. I would like this button to **remove** the checkmark, though, instead of simply dismissing the notification.

If you need any help implementing this, just let me know. We also have a Slack channel, for more real-time discussions. Send an email to dev@loophabits.org if you would like an invitation.

rsri:
Can you please explain the difference between dismissing notification and remove a checkmark? If a habit has been checked, the app doesn't show notification for the same. So dismissing notification implicitly means removing a checkmark right? Am I missing something?

iSoron:
In theory, we never show a notification when there is already a checkmark, so dismissing is the same as removing. In practice, strange things can happen, and it's quite possible that the notification is still showing up, even though a checkmark is present. In these cases, the expected behavior would be for the checkmark to disappear.

In the past, we had a bug where pressing "yes" caused the checkmark to disappear, because we were blindly toggling the checkmark, trusting that no checkmark was not there.

jindalshreyansh:
Please add this feature. I keep the habit notification persistent because I dont want to clear it when I press clear all. But if I fail to do the habit, I have to press check, then open app and uncheck.

iSoron:
This has been implemented and will be available in Loop 1.8.

PR #320, opened 2017-08-04
Implemented cancel to reminders of habits.

rsri:
Code implementation for https://github.com/iSoron/uhabits/issues/319


uhabits-core/src/main/java/org/isoron/uhabits/core/ui/widgets/WidgetBehavior.java
```diff
@@ -61,6 +61,12 @@ public void onRemoveRepetition(@NonNull Habit habit, Timestamp timestamp)
         performToggle(habit, timestamp);
     }
 
+    public void onCancelRepetition(@NonNull Habit habit, Timestamp timestamp)
```
iSoron: Instead of creating a new method that does essentially the same thing as `onRemoveRepetition`, I think it's fine if we just make `onRemoveRepetition` call `notificationTray.cancel(habit)`.

uhabits-android/src/main/java/org/isoron/uhabits/intents/PendingIntentFactory.kt
```diff
@@ -54,6 +54,15 @@ class PendingIntentFactory
                     },
                     FLAG_UPDATE_CURRENT)
 
+    fun cancelNotification(habit: Habit): PendingIntent =
```
iSoron: It is not clear what's the difference between `cancelNotification` and `dismissNotification`. I think it's better to name this `removeRepetition`.

iSoron:
Hi @rsri, I have made a few small changes and merged your pull request. Thank you once again!

PR #324, opened 2017-08-28
Updated Strings Issue #131

towcar:
First time really working with a project in Github, took on a simple issue to make sure I have the process done correct and it seems quite straight forward. Next time though I will claim an issue to work on first. Anyway I fixed three strings pertaining to the deleting of a habit. Super simple stuff.


towcar:
Realized you can delete multiple and hence the wording needs to be change accordingly, I am deleting my pull request for such reason.

PR #326, opened 2017-09-11
Make notifications also blink the LED

llunak:
These two commits implement LED blinking for notifications, similarly to how e.g. calendar apps do it. I've split it into two commits, as I don't know if this is worth having a configuration option (the second commit may be dropped for all I care).


iSoron:
Thank you for the patch, @llunak. I have merged it. If you want to make this feature even better, we could let the user decide the LED color. I don't think it's worth it having them pick the blinking pattern, though.

PR #329, opened 2017-09-12
fix snooze button in notifications

llunak:
The "pref_snooze_interval" preference is manipulated by the ListPreference
class, which according to its docs stores the preference as a string.
Without reverting this part of 864636705d2f43dd54d0045f01d40fb2d65be7cd,
this results in "java.lang.ClassCastException: java.lang.String cannot be
cast to java.lang.Long" when trying to snooze a notification.


iSoron:
Oops, nice catch. Thank you for the patch!

Issue #295, opened 2017-05-23
Allow user to duplicate / clone habits


firelite:
Hi, 

Really loving the app so far! The first thing I felt was missing is the ability to duplicate an existing habit when creating a new habit.  Many of my habits had similar wording so I had to retype each one by memory.  It would be nice to have a quick way to copy existing habits and then modify as needed. 

Thanks!  Wonder if this is something others could use too.

iSoron:
Thank you for the suggestion, @firelite. I agree that this would be an interesting feature.

towcar:
I will take a stride at adding this, shouldn't be too hard. Do we want the edit screen to come up automatically after duplicating?

mr-flibble:
> I will take a stride at adding this, shouldn't be too hard. Do we want the edit screen to come up automatically after duplicating?

I think so, otherwise user will be forced to think which habit is original and which is a copy :)

PR #331, opened 2017-09-13
Allow user to duplicate / clone habits #295

towcar:
This commit should add the feature to duplicate a habit. Doing so brings up the edit/create/duplicate screen allowing the user to make subtle changes to the duplicated habit.

It may need to be updated again as to unselect the original habit. Additionally if you select multiple it only duplicates the first selected. I figure thats probably a good thing since it could get complicated or messy with mass duplicating. Lastly the progress history does not duplicate which I assume makes sense but perhaps users want that duplicated?

Also a last note is I noticed my first commit is also contained into this pull request. Unsure how to remove it but all it does is change some "habits" to "habit", when deleting a habit. Pretty minor, does no damage nor benefit.

### DIFF
--- a/android-base/src/main/java/org/isoron/androidbase/activities/BaseSelectionMenu.java
+++ b/android-base/src/main/java/org/isoron/androidbase/activities/BaseSelectionMenu.java
@@ -40,6 +40,8 @@ public abstract class BaseSelectionMenu
     @Nullable
     private ActionMode actionMode;
 
+    public static boolean duplicateIsSelected = false;
+
     /**
      * Finishes the selection operation.
      */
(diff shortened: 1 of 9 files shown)


iSoron:
Thank you for the pull request, @towcar. This is an important feature that we are missing. Having a look at the code, however, it seems like the checkmarks are not being copied to the new habit. Could you modify your code so that the checkmarks are also duplicated?

Also, please make sure that your code style fits with the existing code style. More specifically, please put the braces in new lines and use wildcard imports.

towcar:
Yep I can definitely make it do that and yeah I forgot to go over my work on the style before submitting. Thanks for the reply!

iSoron:
Closing due to inactivity. If anyone wants to implement this, please feel free to submit a new pull request.

PR #332, opened 2017-09-14
Custom reminder snooze delay

llunak:
Note that this is sort of a work in progress. Although the code seems to work ok and is usable, I'm new not only to the app but to Android/Java in general, and some of the things in these commits feel a bit hackish, so I thought I'd better get some feedback.

These changes implement another option "Always ask" for the snooze delay, resulting in the Snooze button in the notifications opening a time picker dialog, and snoozing the reminder until that time. What should also be implemented but isn't yet is first showing a popup with a list of the preset snooze delays, which would just snooze for the given time, and extra option "Custom" leading to the time picker dialog.

### DIFF
--- a/android-pickers/src/main/java/com/android/datetimepicker/time/TimePickerDialog.java
+++ b/android-pickers/src/main/java/com/android/datetimepicker/time/TimePickerDialog.java
@@ -63,6 +63,7 @@ public class TimePickerDialog extends AppCompatDialogFragment implements OnValue
     private static final int PULSE_ANIMATOR_DELAY = 300;
 
     private OnTimeSetListener mCallback;
+    private DialogInterface.OnDismissListener dismissListener;
 
     private HapticFeedbackController mHapticFeedbackController;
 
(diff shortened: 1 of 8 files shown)


llunak:
Update: I now consider the feature ready.

iSoron:
> These changes implement another option "Always ask" for the snooze delay, resulting in the Snooze button in the notifications opening a time picker dialog, and snoozing the reminder until that time.

That's an interesting feature which could work really well. Thank you for the pull request. I will test it soon and let you know what I think.

iSoron:
Hi @llunak, I have merged your pull request. See e97047387683. Thank you for the patch!

After merging your changes, I made a few modifications, mostly to simplify the code of the SnoozeDelayActivity. In your version of the code, after picking the delay, we were sending an intent to ReminderReceiver, but there is no need to do that. After any activity in the app is started, it has access to all the objects, so we can call the methods of `reminderController` directly.

Other than that, everything was fine. I think this is a great feature. Thank you for suggesting and implementing it!

PR #334, opened 2017-09-18
Fix tests for some locales (e.g. cz_CZ)

llunak:
The CSV fix also fixes the actual CSV export to provide a correct file, not just a test.


iSoron:
Merged. See e052a144bd4b7bf3b9b50658910a0561c0507f31.

PR #337, opened 2017-09-27
fix name of exported files (correct the time)

llunak:
This reverts part of a75a27ad, where this got changed presumably by mistake.


iSoron:
Thank you for the patch!

PR #339, opened 2017-09-28
don't reset habit position when editing it

llunak:
Without this, merely hitting "Save" in the edit dialog reset Habit's position to 0, as it seems such internal data is not kept while editing.


iSoron:
Thank you for the patch! I have also modified our automated tests to make sure that a bug like this is not reintroduced. Have a look at be9c2ff64d107ccfaec085e8629496bc05a524d.

Issue #45, opened 2016-03-20
Widget with multiple habits


sojusnik:
A widget that contains all habits as a list would be very handy to tick habits off, like it's already implemented in [HabitBull](https://play.google.com/store/apps/details?id=com.oristats.habitbull) and shown [here](http://i.imgur.com/CQGFOIn.png).

flowgunso:
The priority on this feature should be raised, it improves usability.

mr-flibble:
I also think that this is an important feature that the application is missing.

yetanothernerdk:
This will be a great feature, agree with @flowgunso, priority should be raised.

vyu1:
Hi all, I haven't contributed to this project before, but is it okay if I try to work on doing this? I've found this feature lacking since first installing the app.

vyu1:
Hi all, I haven't contributed to this project before, but is it okay if I can start working on this? I've found this feature lacking since first installing the app.

iSoron:
Hi @vyu1, pull requests are very welcome. How are you planning to make the checkmark widget support multiple habits? Widgets are very limited on Android, but one thing we could use is the [StackView](http://www.broculos.net/2011/12/android-101-how-to-create-stackview.html) layout. If you have any other ideas, let me know.

vyu1:
@iSoron I was thinking of making a new ListView layout (widgets don't support `RecyclerViews`) that looks similar to the image @sojusnik posted above, http://i.imgur.com/CQGFOIn.png and to `HabitsActivity`.
However, using a `StackView` could also be interesting, and I'm open to anyone's input on which I should try to implement.

woj-tek:
+1

A widget with a view similar to main app, with possibility to select which habits to display (this would go hand in hand with #325)

iSoron:
Thanks to @vyu1, it is now possible to stack multiple widgets together. This works not only for the checkmark widget, but for any other widget we currently have available. Please see the screenshot below.

At this time, I am not planning to add a widget similar to the app's main screen. This is mainly due to Android limitations. I already said this before, but Android widgets are truly limited on what they can do, or even in what they can look like. We already employ a number of hacks to work around these limitations and to make our current widgets look and behave as well as they do. I don't think it would be possible to recreate the main app screen on widgets. Even if we somehow managed to do it, I would expect it to be very sluggish, and have poor usability. That being said, anyone is free to try, of course, and I would be glad to review such a patch.

[...]

PR #346, opened 2017-10-16
StackView Widgets (#45)

vyu1:
I enjoy using the checkmark widget and really like it's transparent design, but having 5-7 checkmark widgets takes up way too much screen space on my home screen. I implemented a StackView style for the existing Checkmark widget. This new feature is related to https://github.com/iSoron/uhabits/issues/45.
To test my changes, place down a checkmark widget and a new `HabitPicker` dialog should appear. This dialog allows you to select up to as many habits as you prefer. If you only select 1, then the app will create the familiar Checkmark widget we have today. If you select multiple, then the app creates a StackView-style Checkmark widget in which you can scroll through the specific habits you selected.
There are a few things to be discussed:
1. All the widgets themes are opaque right now instead of transparent. I'm not really sure how I can change the Context theme for just the StackView-styled Checkmark widget.
2. I left a TODO comment in the code about this one. If the adapter for the StackView is empty or hasn't been retrieved by the OS, an "empty view" appears as a placeholder. Right now, the empty view consists of just a TextView that says, "Checkmark Stack Widget."

I would love any and all feedback!


iSoron:
Hi Victor, thank you for the pull request! I think this is a great feature. The StackView worked even better than I expected. I have been testing this on the emulator and I came across two glitches:

1. When the user clicks a widget, sometimes the entire stackview disappears for a second and is replaced by a loading text. See the screen capture below. Do you know what might be causing this, and have you tried to fix it? I can have a look into it myself if you would like.

2. When the user adds two stacks of checkmark buttons, they both end up with the same cards. Even if the widget is first removed and then re-added, it seems like habits that were selected for the first stack still show up.

device

vyu1:
@iSoron Details of the changes I just pushed:
1. At the time you tested this new feature, I allowed the user to pick which habits they wanted in their stack-view widget. You pointed out an error with this feature (the 2nd glitch you commented above with). I realized that this feature will not be possible. The method `setRemoteAdapter` updates any widgets with the same layout all exactly the same way. If you look in the `RemoteViews` API, you can see that there is a `setRemoteAdapter` method that _can_ set the adapter for widgets individually, but it's deprecated and, if used, will simply ignore the value of the widget id we pass it.
2. I think I made a pretty cool change since last time, which was supporting the "stack view" widget design for _all_ the widget types we have now! To test, place down a home screen widget, click the new "create a stack view widget" button at the top of the dialog that opens up, and you should see a stack view widget for any widget type you choose.
3. Lastly, I have actually spent most of my time trying to fix glitch #1 you mentioned above, but have not been able to fix it. I think the "flash" that we see is because `onDataSetChanged()` is being called in `StackWidgetService`.
Please let me know what you think of these points!

iSoron:
Hi @vyu1, sorry about the long delay. I finally had some time to work on this, and I was able to reduce the flicker problem considerably. It's not perfect, but I think it's good enough. I went ahead and merged your patches into the `dev` branch. Thank you a lot for your help; I'm sure the users will also appreciate it. Also, please let me know if you would like to work on anything else.

vyu1:
Hey @iSoron, sorry I haven't responded to this message in over a month!! I'm so excited to hear this has been merged in and I appreciate your thanks! Hopefully I can find another thing to work on and contribute to, thanks again 😄

Issue #338, opened 2017-09-28
Reminders do not trigger if app is not running


llunak:
See https://github.com/llunak/uhabits/commit/2bc20255c0444e3f50f62243efdb89ba82b9eda0 for the story.

What I'm not sure about is how to fix this properly. The commit above is simple and works, but it seems somewhat pointless to load habits in the background if the main thread very quickly blocks on it somewhere (setting up reminders does this on app start). In other words, the dev version of the app simply blocks the UI during the whole startup (which 1.7.6 doesn't AFAICT), so maybe handling this correctly requires going a level up and thinking about the design of how it should work.

PR #348, opened 2017-10-18
synchronize HabitList properly

llunak:
When the app is not running and ReminderReceiver receives ACTION_SHOW_REMINDER,
the (SQLite)HabitList will start loading its data in the background for some
reason, and if the loading takes a moment, ReminderReceiver will call
the unsynchronized HabitList.getById() before the loading finishes,
failing to find the habit for which to show the notification.

I originally submitted this only as https://github.com/iSoron/uhabits/issues/338 , but since this commit fixes a problem and I don't see a regression compared to dev, I've simply created a pull request and the other stuff can be handled later.

### DIFF
--- a/uhabits-core/src/main/java/org/isoron/uhabits/core/models/memory/MemoryHabitList.java
+++ b/uhabits-core/src/main/java/org/isoron/uhabits/core/models/memory/MemoryHabitList.java
@@ -155,14 +155,14 @@ public class MemoryHabitList extends HabitList
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


iSoron:
Thank you for the patch! Regardless of that bug, these methods should be `synchronized` in any case, otherwise the classes are clearly not thread-safe. There are probably other bugs out there related to this. I will see if there's anything we can do about blocking the UI thread on ReminderReceiver.

PR #355, opened 2017-11-05
Save snoozed reminders

llunak:
If a reminder is snoozed, the app may actually discard the snoozed reminder and not show the reminder again, for a number of reasons such as:
- Any (even different) habit is edited, which results in ReminderScheduler.onCommandExecuted() to simply do scheduleAll(), scheduling the snoozed habit at its configured time instead.
- The app is (re)started (be it after version upgrade, or simply because it's no longer running).

These patches implement saving of such snoozed reminders, so that they are not forgotten until they are activated again or discarded.

Note that these changes depend on https://github.com/iSoron/uhabits/pull/332 (included here as the first 5 commits).

### DIFF
--- a/android-pickers/src/main/java/com/android/datetimepicker/time/TimePickerDialog.java
+++ b/android-pickers/src/main/java/com/android/datetimepicker/time/TimePickerDialog.java
@@ -63,6 +63,7 @@ public class TimePickerDialog extends AppCompatDialogFragment implements OnValue
     private static final int PULSE_ANIMATOR_DELAY = 300;
 
     private OnTimeSetListener mCallback;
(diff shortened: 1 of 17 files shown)


iSoron:
@llunak Replying to you a year and a half later! The reason I never merged this pull request is because it was clear that Android was going to implement their own version of notification snoozing. Indeed, this feature became available on Android Oreo, so we no longer need to roll out our own. I can't say I'm satisfied with the user interface they decided to use, but I think it's pointless for us to replicate something that the OS can already do. In any case, I shouldn't have left you hanging for so long. I'm terribly sorry for that.

PR #370, opened 2018-01-11
Fix typo Tr translation

lacivert:


matkoniecz:
Note that readme has

> Translate the app into your own language. If you are not a native English speaker, and would like to see the app translated into your own language, please join our open translation project. If the translation is already completed, you are also very welcome to join and proofread it.

linking to http://translate.loophabits.org/

lacivert:
@matkoniecz thank you. There is already a right translation there for 3 months. Anyway, that is not something very important.

matkoniecz:
In that case this PR probably should be closed (there should be a button on the bottom, below comment box).

PR #405, opened 2018-05-02
Fixed plurals problem with habit(s) deletion

recheej:
I'm back after a long absence :). Look at this PR: #260 

I implemented your changes. Only one problem. Look at ListHabitsScreen.kt line 106. There I have to test if it's a delete habits command because the showMessage method takes in a plural resource and the getExecuteString method doesn't. Can you think of a better way of doing this?


recheej:
Also, someone will have to go back and make these plural changes for the other languages :)

iSoron:
Thank you for the pull request, @recheej. I was also away from the project for a while. The reason I am not merging this pull request right now is that, as you mention, we would need to go back and update all strings. I am actually considering removing the ability of deleting multiple habits at a time, since this is a highly destructive operation, in which case we wouldn't need to change the messages.

Issue #42, opened 2016-03-19
Track numerical habits


iSoron:
It would be nice if Loop could track numerical habits, in addition to yes/no habits. It could be useful, for example, for tracking: number of repetitions of a certain exercise; body weight or fat percentage; number of minutes listening to, or reading books, in a foreign language; and many other examples.

Development Subtasks:
- [x] Modify database scheme and model classes
- [x] Update computation of scores
	- [x] Daily target
	- [x] Weekly/monthly target
- [x] Implement dialog for creating/modifying numerical habits
- [x] Improve dialog for selecting what kind of habit to create
- [x] Display numerical data on the main screen
- [x] Update detail screen
	- [x] Show target in subtitle card
	- [x] Create new bar plot chart
	- [x] Allow user to change values in calendar chart (#627)
	- [x] Hide streaks card
- [x] Update notifications
	- [x] Replace Yes/No buttons by "Enter" (#627)
- [x] Update widgets
	- [x] Show numerical habits in checkmark widget
	- [x] Hide numerical habits when creating streak widgets
    - [x] Create widget for target chart
- [x] Update Tasker plugin
	- [x] Hide numerical habits from existing check/uncheck/toggle actions
	- [x] Add action to increment/decrement numerical value

Donaira:
Users can set a goal to read X pages, exercise for Y minutes, do Z push ups, etc. Then the question would be "Did you read X pages today?" Therefore, as a user, I don't think this feature is necessary. This app is for developing good habits, not for logging your workouts. IMHO. :)

Guitar1st:
This function can be used in a different way. I found the idea of it in the book "Triggers". The author suggests to ask yourself "Did I do everything possible to reach the goal?". And the answer could be a number from 1 to 10. It is the measure of person's diligence in the current day.
I really miss this feature

woj-tek:
I for one would welcome the feature - I'm in the process of doing small tasks (a couple of minutes every hour) and having ability to mark how many I've done on a given day would definitely help with making this a habit.

iSoron:
I have started working on this feature in the branch `feature/numerical-habits`. Please, be advised that the code is very experimental, and may corrupt your database. I am currently focusing on habits that can be measured by numbers, such as:

- How many...
    - km/miles did you run today?
    - push-ups/reps did you do?
    - glasses of water did you drink?
    - calories did you eat?
    - pages did you read?
    - words did you write?
    - math exercises did you solve?
[...]

PR #419, opened 2018-06-07
Numeric Goals/Type Selector and Increment Gesture

jakedp7:
Not sure if you're taking PRs for the hidden numeric goals functionality, but I made some modifications for my own use and wanted to at least let the maintainers decide.

Summary of changes:
- Got rid of the habit type dialog upon habit creation, to clean up the interface a bit. Instead, the habit type is selected by a spinner in the habit editor dialog. 
- Implemented GestureDetector in NumberButtonView. A single tap increments the count of a numeric habit, a double tap decrements it, and press-and-hold works as it did (brings up the value selection dialog).


iSoron:
Thank you for the pull request, @jakedp7. I will have a closer look, but, to justify why the app is currently the way it is:

i) As discussed in #42, I am planning to include other types of numerical/categorical habits. At some point, we will need to explain to the user more clearly how do these types differ, and perhaps show them examples of each habit. I was planning to do this in the habit type dialog. Besides, adding new habits is not something that users will do every day, so I am okay with this operation having some extra steps.

ii) If the user has activated the "toggle habits with short press" option, then a single tap should bring up the value selection dialog. I'm aware that it would be very useful to have a quick shortcut to increment/decrement the value, but we also need to worry about accidental data entry. I'm not sure what the best solution would be.

jakedp7:
Alright @iSoron, thanks for the feedback - I'll go ahead and close this then.

iSoron:
I agree that #42 is too broad. I am planning to create more specific issues soon. In the meanwhile, these are some areas I'd like to see improved before numerical habits become available to all users:

- [ ] Make widgets work with numerical habits (in particular, the CheckmarkWidget)
- [ ] Make notifications actions work with numerical habits
- [ ] Fix streak calculation (number of consecutive days above the threshold)
- [ ] Implement conversion between types (yes/no to numerical, and vice-versa)
- [ ] Update the tasker plugin

Let me know if you would like to work on anything here, or if you have any questions.

jakedp7:
@iSoron I'd love to help. Is there a better venue to talk about it in? I'm not a part of the Slack mentioned in the Wiki.

iSoron:
@jakedp7 The Slack channel was not very active, and it was behind closed doors, so I shut it down. If you would like to discuss a new feature, please feel free to open an issue or submit a pull request. In this way, others can join the discussion too.

Issue #442, opened 2018-09-30
No "Delay" option (snooze) on notifications


aaviran:
Hi, Recently installed the app on my phone and there is no "Delay" option on the notifications. It clearly says in the app description: 

> Easily check, dismiss or snooze your habit directly from the notification, without opening the app.

So I'm pretty sure this is a bug. My phone is a Sony Xperia XZ2 Compact. I can provide more details if needed.

My loop app version is 1.7.9, Android version 8.0.0

Thanks

aaviran:
![screenshot_20180930-211007](https://user-images.githubusercontent.com/10272703/46260975-c8d16d80-c4f5-11e8-87d6-63751420d2a6.png)

MarosPataky:
Hi @iSoron, I provided a PR for this. Please, check it out when you have time.

Erudition:
You're missing a major detail. In modern android there's snoozing built in, so the snoozing provided by the app gets disabled. You didn't specify which one wasn't working for you.

Your notifications look exactly like mine, and mine snooze just fine, so that screenshot isn't helpful.

In case you didn't know how you can snooze any notification in Oreo:


Begin swiping the notification away, but stop short once the underlying action buttons are uncovered. In the case of TouchWiz, the snooze action button is a bell. You should be able to figure it out from there.

iSoron:
As @Erudition explained, this feature has been removed since, starting from Android Oreo, you can already snooze the notifications of any app by dragging them to the right. Furthermore, our previous implementation did not work reliably on some phones.

PR #444, opened 2018-10-08
Removed API_VERSION < 26 condition for snooze action to be displayed.

MarosPataky:
I started working on this based on https://github.com/iSoron/uhabits/issues/442.

I removed this conditional, since I was not able to see Snooze ("Later") button on my device with app version 1.7.9. My device is Xiaomi Mi A1 running Android v. 8.1.0.

The condition removed did not make any sense to me and fixed the issue.


srujanb:
I think keeping the snooze option does not hurt at all. I really want that even when I'm using Android O. The reason is, default Android functionality is very limited. I have "sticky" notification in which I cannot dismiss it unless I "check" it. If I do not want to do the task on a particular day, I just have to let the notification stay there forever. Please being that option for O too.

Issue #51, opened 2016-03-20
Display combined statistics for groups of habits


sojusnik:
To have the statistics (success rate, current streak and longest streak of a habit) at a glance of all habits would be very helpful.

iSoron:
Thank you for the suggestion. This is somewhat related to issue #31.

SuzanaK:
It would also be great to have the overall statistics in a widget.

ysagal:
This would be particularly useful for a widget to be able to see total progress across all habits at a quick glance.

nj999:
Hello.

I want to add my comments to this thread. We would like to see a combined score/points for one day and track this day over day. Like in google fit you can see that today you have walked 5000 steps and yesterday 4500 steps. So this way we can track progress of all of our goals together. Not sure if this is a new feature or connected to the same comment from ysagal about total progress across all habits. 

But please consider this request/enhancement.

chreekat:
OP didn't explicitly mention "Habit strength", which is the widget/chart I would most like to see. Of course, it may be rather confusing unless one color-codes their habits, since there would be virtually no room for labels.

zaheernathani:
When can we expect this.

chennemann:
@zaheernathani Could you please elaborate on how that could be useful. I have more than 25 habits and seeing them in one graph would be everything but confusing and unorganized.
Is there a specific use-case you need that aggregated view for?

Also regarding the timeline. Since this is an open-source project there is no way to provide you with an ETA. If you want/need that feature you are welcome to submit a PR that implements it.

PR #446, opened 2018-10-27
Allow multiple habits on strength chart

jknoxville:
This is a start at tackling one of the items in #51 
Displaying multiple habits on the same strength chart.

Now when creating a widget from several habits, if you choose "Habit
strength" as the type, then instead of stacking, they will be displayed
on the same chart.

It's a bit hacky, and needs cleaning up some, so I'm just looking for a high level review at the moment. If you think it's in the right direction I can carry on. Otherwise I'd be interested to hear how you would approach it.

Known issues:
* All references to Habit were removed from ScoreChart in a refactor, so maybe we'll want to do this differently, as I've just added them back to get it working quickly.
* The legend on the graph doesn't wrap around.
* The legend text can be hard to read depending on the colors of your habits. It might be a better idea to automatically assign colors to each habit just for this chart, this would also ensure the lines are all different colors, which is a current issue. Or maybe change the background for it or something, I'm no good at UI design...
* The way I've used ArrayLists everywhere feels a bit hacky.
* I broke the populateWithRandomData method for convenience.
* I also haven't touched or even ran the tests.

Screenshot:
screen shot 2018-10-28 at 00 49 29

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/views/ScoreChart.java
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/views/ScoreChart.java
@@ -21,7 +21,6 @@ package org.isoron.uhabits.activities.common.views;
 
 import android.content.*;
 import android.graphics.*;
-import android.support.annotation.*;
 import android.util.*;
 
 import org.isoron.androidbase.utils.*;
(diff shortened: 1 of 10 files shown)


MohamedKiouaz:
I don't have the level to read the code but I would appreciate this feature.

iSoron:
Thank you for the pull request, @jknoxville! I will have a look and get back to you soon.

iSoron:
If you are still working on this, please let me know.

Issue #432, opened 2018-08-16
Unable to Read the Number(s) in Best Streaks


iknownothingaboutcomputers:
Using dark theme. If the streak is short, then the colour of the number and the background is the same, thus making it difficult to read the number. If the streak if long, then the background colour changes to whatever colour user has set for that particular habit.

iSoron:
Thank you for the bug report. Could you please upload a screenshot, so the problem becomes more clear?

iknownothingaboutcomputers:
@iSoron 

![screenshot_20180826-143656](https://user-images.githubusercontent.com/26545454/44856860-07a5a680-ac5e-11e8-90ac-f83f0c4b6d1a.png)

iknownothingaboutcomputers:
@iSoron Not easy to see those numbers when viewed on a phone with screen brightness set to minimum.

chennemann:
This also applies to the other color modes and the calendar chart as well. I fixed that in the referenced pull request. There I also attached screenshots of every theme configuration. Light, dark and black.

PR #460, opened 2019-01-20
feat: Improve readability of history and streak charts

chennemann:
References: #432 

I improved the readability of the dates and the streak count.

image
image
image

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/views/HistoryChart.java
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/views/HistoryChart.java
@@ -73,13 +73,13 @@ public class HistoryChart extends ScrollableChart
 
     private int colors[];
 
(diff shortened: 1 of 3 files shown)


chennemann:
Do you mean only for the light theme or also for the others? 

When I remember correctly in the dark/black theme the numbers on the implicit days were not very visible. 

You are right for the light theme though.

iSoron:
Thank you once again for you pull request, @chennemann. I am not merging this one since, as explained in #486, we will eventually replace the Android implementation with [the code in the shared module](https://github.com/iSoron/uhabits/tree/082c575f82479dbfa5ff418c46b2cf4cff94f40e/core/src/commonMain/kotlin/org/isoron/uhabits/components). As you can see, the CalendarChart in that module already takes into account the color of the box to decide what text color to use. I have not converted the other charts yet. If you would like to work on that, you are very welcome.

Issue #267, opened 2017-03-23
Create a landing page for the project


iSoron:
Our domain  [loophabits.org](http://loophabits.org/) currently shows a blank screen. It would be nice to have a simple landing page for the app, containing some descriptions, screenshots and links to Google Play / Github. The descriptions can be copied from the [Google Play page](https://play.google.com/store/apps/details?id=org.isoron.uhabits). A high resolution version of the screenshots is available [in our repository](https://github.com/iSoron/uhabits/tree/dev/screenshots/original). 

Since I am hosting this on GitHub Pages, it has to be a static website.

nickdbush:
I'd love to help with this! Would that be OK? I'm guessing you'd like to see designs fist?

iSoron:
Hi Nick, it would be great if you could help! If anyone else wants to post alternative designs too, so we can try different ideas, please feel free.

atchyut-re:
Hello @iSoron, I have recently started using your app. I was very impressed with the features it have.
After that I had noticed its an open source project and wished to add my contribution.
Being a web developer I found this feature which I can do. Can you please let me know if still available to develop?

atchyutn:
Hello @iSoron, I have recently started using your app. I was very impressed with the features it have.
After that I had noticed its an open source project and wished to add my contribution.
Being a web developer I found this feature which I can do. Can you please let me know if still available to develop?

nickdbush:
I don't mind, as I've got quite a lot going on at the moment. Here's where I'd got up to design wise until I had to stop:

![loop](https://cloud.githubusercontent.com/assets/10998201/26150090/c749e442-3af4-11e7-90b4-6a41b0ce8228.png)

If you'd like to work on this together @atchyutn, drop me an email

atchyutn:
Hello @nickdbush, This so cool! :ok_hand: 
I'm not an expert but a beginner, I can help you along if its okay for you. I wish to share some of my ideas.

Will drop an email with my thoughts and we can move further.

Thank you for the opportunity. :smile:

padamczewski:
I really like the app and would like to contribute as well :smile: 

@nickdbush @atchyutn: What's the status of the stuff that you were working on? It would be good to see some code to get a better idea of what you wanna achieve.

RayBB:
@nickdbush have you worked on this at all?

I'm willing to build this but would like some guidance on the design. I could make a simple page as shown in the mockups.
@iSoron any thoughts?

[...]

PR #461, opened 2019-01-20
feat: Rudimentary implementation of a landing page

chennemann:
References: #267 

This really is a rudimentary implementation of the proposed design. It does not follow any coding best practices but it is at least something until a good page is implemented.


iSoron:
Thank you for the pull request, @chennemann, but I prefer to leave it blank until we have a completed website.

Issue #463, opened 2019-01-20
Break days on non-daily habits should not decrease the strength


chennemann:
I think it's a bit counter intuitive and even demotivating that the strength goes down for days on which you don't have to complete the habit (let's call them break-days).

It's best explained with a picture:  
![](https://i.imgur.com/GoOBZCX.jpg)

As you can see the habit is completed every time it should be completed and the streak is ongoing. Nevertheless I have to rebuild the strength every week again because it goes down on Saturdays and Sundays. 

I don't know how the algorithm currently works but I would propose that break-days should neither increase nor decrease the strength of a habit. 

If my proposed solution is desirable I could work on implementing this change. So please let me know if I should start working on that.

chennemann:
I just implemented a working change. But I'm actually not happy with my approach. 
It works for my use case but it is quite inflexible so we should look for a different approach. Maybe someone has an idea that would diminish the nasty double check and that also works with numerical habits.

Anshoe:
Agreed

jimfarrand:
I built the current dev branch and it looks like the behaviour is a little different from the version in the play store.  With regards to the score calculation, implicit ticks for break days are considered NOT DONE on the Play Store version, but are considered DONE on the dev version.

The dev version behaviour seems a lot more sensible to me than the Play Store version.  However, it's not quite what @chennemann originally proposed, which is that break days should neither increase nor decrease the score.  That said, although I prefer the original suggestion, I could live with the behaviour in the dev branch.

Is it worth me implementing the change as originally requested, or is there now consensus that the dev behaviour is correct?  I'm happy to try to make the change if it will be accepted.  Otherwise maybe we should close this issue?

chennemann:
maybe you can have a look at my current implementation that I referenced in this issue and you find a more flexible solution than I did. I actually don't like both versions. neither increasing nor decreasing should be the actual behavior because you neither work on building your habit nor are you neglecting it.

jimfarrand:
Work in progress.  Tests need to be updated.

The interface to Checkmark and friends is kinda... not good.  It's not really providing any encapsulation and so callers to getValue() and friends have to be hyper-aware of the implementation details in other parts of the code.

But fixing that looks like a real bug job that is going to break anyone else that has WIP so I don't think I should tackle it in the scope of this issue.

jimfarrand:
[...]

PR #464, opened 2019-01-22
Feat/ignore break days for score calculation

chennemann:
References #463


iSoron:
I am closing this pull request until we have figured out how to handle this better. Let's keep the discussion at #463.

PR #470, opened 2019-01-23
Feat/always hide archive action

chennemann:


chennemann:
Sorry. That was the wrong branch.

Issue #465, opened 2019-01-23
idea: Usability improvement


chennemann:
I think when multiple habits are selected we should not show the archive action. I would rather suggest that we only hide the edit icon. That is because the actions change positions which is confusing for the user. We should minimize the possibilities of getting to a result and for archiving there are currently two ways depending on the selected habit count. 

I hope it is clear what I mean. If not I can further elaborate.

PR #471, opened 2019-01-23
feat: Always hide archive/unarchive icon

chennemann:
- Improves usability since the icons do not
    jump around when more than one habit is
    selected

References: #465


iSoron:
Thanks, I think this is a good and simple solution.

PR #475, opened 2019-01-27
feat: Change order of calendar and history"

chennemann:
I think it would be better to change the order of the calendar and the history widget. That is because the user is more likely to find out on which days he completed the habit than how many times he did the habit in a certain month.

Also it would put the history and the best streak chart together which kinda belong together in my opinion.

What are your thoughts about this?


iSoron:
@chennemann Thank you for the pull request. Could you please remove the unrelated changes to the gradle file and fix the conflicts? I will merge it.

iSoron:
If you're still working on this, let me know and I will reopen it.

Issue #413, opened 2018-05-28
The contrast of greyed out habits is too low in night mode


hiqua:
I'm one of those users who use night mode for everything, at any time of the day.

Unfortunately Loop Habit Tracker seems to keep the same grey for e.g. habits that have not been completed yet.

This makes it hard to distinguish between those that have been completed in the previous days (grey tick), and those that should be completed (grey cross).

Please consider using a different colour in night mode, to make the distinction easier.

Thanks for Loop Habit Tracker!

PR #476, opened 2019-01-27
feat: Increase contrast for habit overview

chennemann:
- Increased the contrast of the lowContrastText color of the
      black and the dark mode

References: #413


iSoron:
Closing this for the same reason I am closing #460.

PR #480, opened 2019-01-30
Fix background of popup elements in black theme

chennemann:
Popup elements in the pure black mode use the default theming of the dark mode which looks a bit odd because it's too light. So I suppose we should change the color accordingly.


iSoron:
@chennemann Thank you for the pull request. Could you please include a screenshot of the changes? Also, in the future, there's no need to create an issue if you are submitting a pull request at the same time.

iSoron:
If you're still working on this, let me know and I will reopen it.

PR #498, opened 2019-07-23
Changed habit type selection to dropdown instead of popup message

Tthecreator:
Hi there!

I've been working on issue #42 and have a proposal to ' Improve [the] dialog for selecting what kind of habit to create'.

I have done this by changing the old popup message to a dropdown menu.
If this change is accepted I'll first remove/cleanup the old popup message code.
I've also done some extra work to make sure the checkbox to enable/disable numeric habits in settings works without restarting the app.

Right now I'm asking for some feedback on my code, since I'm not that familiar with this kind of android development I had to figure out a Kotlin, how dagger works and I had to take a good look at the design patterns used. I'm not sure I've used them correctly. In my code I've made ListHabitsMenu an Preferences.Listener. This way it can enable/disable the dropdown menu when this feature is enabled/disabled. This way all UI can be in the original list_habits.xml file and once the NumericalHabitsFeatureEnabled setting becomes deprecated it's easily removed, resulting in cleaner code. The problem that I have with this solution is that it's not very nice that ListHabitsMenu knows and does as much as it's doing. Choosing if the menu should be shown or not should be in the uhabits-core. I have no idea how to do that however. Because dagger uses a directed acyclic graph I can't give the behavior a reference to the menu because the menu already has a ListHabitsMenuBehavior.

### DIFF
--- a/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsMenu.kt
+++ b/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsMenu.kt
@@ -73,6 +73,16 @@ class ListHabitsMenu @Inject constructor(
                 return true
             }
 
+            R.id.actionCreateBooleanHabit -> {
+                behavior.onCreateBooleanHabit()
+                return true
+            }
+
+            R.id.actionCreateNumeralHabit -> {
+                behavior.onCreateNumericalHabit()
+                return true
+            }
+
             R.id.actionFAQ -> {
                 behavior.onViewFAQ()
                 return true
--- a/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsMenu.kt
+++ b/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsMenu.kt
@@ -46,30 +46,15 @@ class ListHabitsMenu @Inject constructor(
         hideArchivedItem.isChecked = !preferences.showArchived
         hideCompletedItem.isChecked = !preferences.showCompleted
 
-
         topBarMenu = menu
-        //the habit creation menu should be disabled when numeric habits are also disabled
-        if (!preferences.isNumericalHabitsFeatureEnabled) {
-            setCreateHabitMenuEnabled(false, menu)
-        }
         //let the class add itself as listener
         preferences.addListener(this)
-    }
-
-    override fun onNumericalHabitsFeatureChanged() {
-        if(topBarMenu==null){return}
-        setCreateHabitMenuEnabled(preferences.isNumericalHabitsFeatureEnabled, topBarMenu)
     }
 
     override fun onItemSelected(item: MenuItem): Boolean {
         when (item.itemId) {
             R.id.actionToggleNightMode -> {
                 behavior.onToggleNightMode()
-                return true
-            }
-
-            R.id.actionAdd -> {
-                behavior.onCreateHabit()
                 return true
             }
 
@@ -138,6 +123,7 @@ class ListHabitsMenu @Inject constructor(
      * @param enabled whether the create habit menu should be enabled or disabled
      * @param menu a reference to the menu on which should be enabled or disabled
      */
+    @Deprecated(message = "This function was used to enable/disable the habit creation drop down menu, but since the feature is mostly implemented this is no longer necessary.")
     fun setCreateHabitMenuEnabled(enabled: Boolean, menu: Menu) {
         val habitCreationMenu = menu.findItem(R.id.actionAdd).subMenu
         for (itemIndex: Int in 0 until habitCreationMenu.size()) {
--- a/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsMenu.kt
+++ b/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsMenu.kt
@@ -33,7 +33,7 @@ class ListHabitsMenu @Inject constructor(
         private val preferences: Preferences,
         private val themeSwitcher: ThemeSwitcher,
         private val behavior: ListHabitsMenuBehavior
-) : BaseMenu(activity), Preferences.Listener {
+) : BaseMenu(activity){
 
     override fun onCreate(menu: Menu) {
         val nightModeItem = menu.findItem(R.id.actionToggleNightMode)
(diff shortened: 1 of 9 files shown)


android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsMenu.kt
```diff
@@ -33,15 +33,32 @@ class ListHabitsMenu @Inject constructor(
...
         val nightModeItem = menu.findItem(R.id.actionToggleNightMode)
         val hideArchivedItem = menu.findItem(R.id.actionHideArchived)
         val hideCompletedItem = menu.findItem(R.id.actionHideCompleted)
+
         nightModeItem.isChecked = themeSwitcher.isNightMode
         hideArchivedItem.isChecked = !preferences.showArchived
         hideCompletedItem.isChecked = !preferences.showCompleted
+
+
+        topBarMenu = menu
```
Tthecreator: This is ugly. How could I have done this better. What other way can I get a reference in onNumericalHabitsFeatureChanged?

chennemann:
thanks for your contribution. could you maybe add some screenshots from before and after? unfortunately I don't have time right now to review your code but I try to do so soon.

Tthecreator:
Hi here are the screenshots you requested:
Before:
Screenshot_20190801-122039
After:
Screenshot_20190731-154908

chennemann:
I definitely like the change. the only thing that could be improved would be the labels (not the scope here). I'll try to have a look at the code later.

iSoron:
Thank you for the pull request, @Tthecreator. The reason the original version has a popup instead of a dropdown is that I was planning to include some extra information, explaining to the user what different types of habits are, and maybe some examples. This is important because, in the future we will have multiple types of numerical habits. In any case, we can proceed with the popup for now and work on an improved dialog at a later time.

android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsMenu.kt
```diff
@@ -33,15 +33,32 @@ class ListHabitsMenu @Inject constructor(
...
+        //the habit creation menu should be disabled when numeric habits are also disabled
+        if (!preferences.isNumericalHabitsFeatureEnabled) {
```
iSoron: Since this numerical habits feature is almost complete, I think it's fine if we get rid of the feature switch. That is, let's just make the dialog available at all times. This will simplify your code quite a bit. Please feel free to delete the code to activate/deactivate this feature in the preferences screen.

android/uhabits-android/src/main/res/menu/list_habits.xml
```diff
@@ -38,31 +51,31 @@
                 android:id="@+id/actionHideArchived"
                 android:checkable="true"
                 android:enabled="true"
-                android:title="@string/hide_archived"/>
```
iSoron: Please avoid proposing changes that only modify white space.

Tthecreator:
> Thank you for the pull request, @Tthecreator. The reason the original version has a popup instead of a dropdown is that I was planning to include some extra information, explaining to the user what different types of habits are, and maybe some examples. This is important because, in the future we will have multiple types of numerical habits. In any case, we can proceed with the popup for now and work on an improved dialog at a later time.

Maybe we can explain in the intro tutorial? Or add an 'help' option to the drop down menu?
I'll make the requested changes tonight, thank you for your feedback.

android/uhabits-android/src/main/res/menu/list_habits.xml
Tthecreator: Sorry this one slipped by me, and when reviewing my own code I couldn't be bothered to revert it. But if you want it I'll do it.

android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsMenu.kt
```diff
@@ -107,5 +119,19 @@ class ListHabitsMenu @Inject constructor(
...
+    @Deprecated(message = "This function was used to enable/disable the habit creation drop down menu, but since the feature is mostly implemented this is no longer necessary.")
+    fun setCreateHabitMenuEnabled(enabled: Boolean, menu: Menu) {
+        val habitCreationMenu = menu.findItem(R.id.actionAdd).subMenu
+        for (itemIndex: Int in 0 until habitCreationMenu.size()) {
+            val menuItem = habitCreationMenu.getItem(itemIndex)
+            menuItem.isEnabled = enabled
+            menuItem.isVisible = enabled
+        }
+    }
+
```
iSoron: If this code is no longer needed, please remove it.

android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsMenu.kt
```diff
@@ -33,15 +33,22 @@ class ListHabitsMenu @Inject constructor(
...
         val hideArchivedItem = menu.findItem(R.id.actionHideArchived)
         val hideCompletedItem = menu.findItem(R.id.actionHideCompleted)
+
         nightModeItem.isChecked = themeSwitcher.isNightMode
         hideArchivedItem.isChecked = !preferences.showArchived
         hideCompletedItem.isChecked = !preferences.showCompleted
+
+        topBarMenu = menu
+        //let the class add itself as listener
+        preferences.addListener(this)
```
iSoron: Why do we still need to listen to preference changes here?

android/uhabits-core/src/main/java/org/isoron/uhabits/core/preferences/Preferences.java
```diff
@@ -346,6 +336,10 @@ default void onNotificationsChanged()
         default void onSyncFeatureChanged()
         {
         }
+
+        default void onNumericalHabitsFeatureChanged()
+        {
+        }
```
iSoron: It seems like this method is longer needed. Please remove it.

android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsMenu.kt
```diff
@@ -33,15 +33,22 @@ class ListHabitsMenu @Inject constructor(
         private val preferences: Preferences,
         private val themeSwitcher: ThemeSwitcher,
         private val behavior: ListHabitsMenuBehavior
-) : BaseMenu(activity) {
+) : BaseMenu(activity), Preferences.Listener {
+
+    private lateinit var topBarMenu: Menu
```
iSoron: This field is no longer needed. Please remove it.

Tthecreator:
Thank you for your rigorous review iSoron!
I've made the changes you requested and hope it is good enough to be merged now.
I'll merge these changes onto #499 when this issue gets merged.

PR #524, opened 2019-11-09
Preserve selections when reopening EditSettingActivity

mickdekkers:
I noticed that when I opened Tasker to edit my settings for the Loop plugin, the spinners weren't set to the habit/action I selected previously. This fixes that.

> The host may also pass the Bundle back to the Activity in the future, if the user wishes to edit the plug-in's configuration.

Source: https://www.twofortyfouram.com/developer/


iSoron:
Thank you for the pull request, @mickdekkers.

PR #543, opened 2020-01-05
Migrate to AndroidX

TacoTheDank:
Yeah...

Edit: Also, check these for any changes you might have to make (or features to implement) for the upgrade from 1.0.0 to 1.1.0.
https://developer.android.com/jetpack/androidx/releases/appcompat
https://developer.android.com/jetpack/androidx/releases/preference
https://developer.android.com/jetpack/androidx/releases/test (this one for the test modules)

No errors, apart from the same ones from my previous pull request.


TacoTheDank:
Btw, would you mind if I made another pull request that removes all the *s from all the imports? I had to check through all 189 changed files in this pull request for import errors (because of migration errors due to the imports using * instead of classes).

TacoTheDank:
https://github.com/iSoron/uhabits/pull/544

PR #547, opened 2020-01-08
Add Notes to Habits.

### LINKED ISSUE #56: Allow user to add notes to habits
Currently, habits have titles and questions for the reminders. It would be nice to have a general "notes" field to allow the user to store whatever other information they would like about the habit, such as the reasons. These notes would be optional, and displayed only on the top of the statistics page.

This covers issues #54 and #47.

### LINKED ISSUE #529: Add a notes field to habits which is not shown in notifications for privacy reasons
Habits have a name and optional question currently. I would propose to add a "notes" field for two reasons:

* Both name and question are shown in notifications. "Notes" should not be shown outside of the app. This allows more private notes to be entered there instead what is shown in your notifications as they are far more likely to be (accidentally) shown to someone else when you show something on your phone. For this case, you could move the habit "requirements" (question field) to the notes and only use a "codename" in the name field that is clear to you but nobody else. This is related to #387 but no duplicate.

* Allow to enter more details of the habit (not privacy related). I am not sure if that reason counts because you could just enter it all into the question field. At least for me, only the first line of the question field is shown in notifications.

Workaround: Enter ".          " some non-space character and a lot of spaces before your actual question to move the actual private question out of the screen.

recheej:
Resolves #56 and resolves #529 

image

image

**Changes**

* added ```question``` column to ```Habit``` table. This is a nullable text column, similar to description.
* added migration from db 22 -> 23. What this does is copy our old "description" to the question column and set the description column to null.
* Refactor ```edit_habit_name``` layout to a constraint layout (added dependency). Related to this is updating ```NameDescriptionPanel``` (the view class that uses ```edit_habit_name```) to include the notes input.
* I think most everything else is just updating code across the app to refer to the ```question``` field instead of ```description```

**Follow up questions/comments for @iSoron** : 

1. where/how do you actually want this new field to be shown? :)
2. Right now importing from various importers takes their "description" field and puts it into uhabits "question" field since I just replaced the name. Maybe that doesn't make sense anymore and it should go into the new notes/description field?
2. I ignored one test. It's failing when trying migrate 21 to 22 because of the ```findAll``` method. I think the method ```findAll``` should be refactored. The reason the test is failing is because ```findAll``` queries all columns on the ```HabitRecord``` class even though version 22 didn't have that column. Maybe I'm over thinking things here. Let me know if that makes sense. The failure reported by the unit test shouldn't actually happen in prod.


iSoron:
Thank you for the pull request! I will review the code changes in more detail soon. To answer your questions:

1. I think the best place to show it is as a new card in the ShowHabitActivity, either directly below the current SubtitleCard, or after the OverviewCard. Please feel free to try other ideas. When no notes are entered, the card could disappear completely.
Screenshot_1578482369

2. Yes, I think the importers should populate the new `description` field instead of question.

3. I agree that test is invalid. Please feel free to either remove it or rewrite it. Before running `helper.migrateTo(int)`, we cannot assume that the current source code can understand the (old) database format. Entire tables might be gone.

A couple of other comments:

4. Can you please make sure that the notes/description are also exported when we generate the CSV files? The relevant method is `HabitList.writeCSV`

5. I am getting a crash when I try to edit a habit created in a previous version of the app:

	    java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.isEmpty()' on a null object reference
	        at org.isoron.uhabits.activities.habits.edit.views.ExampleEditText.updateText(ExampleEditText.java:102)
	        at org.isoron.uhabits.activities.habits.edit.views.ExampleEditText.setRealText(ExampleEditText.java:89)
	        at org.isoron.uhabits.activities.habits.edit.views.NameDescriptionPanel.populateFrom(NameDescriptionPanel.java:109)
	        at org.isoron.uhabits.activities.habits.edit.EditHabitDialog.populateForm(EditHabitDialog.java:229)
	        at org.isoron.uhabits.activities.habits.edit.EditHabitDialog.onCreateView(EditHabitDialog.java:112)
	        at androidx.fragment.app.Fragment.performCreateView(Fragment.java:2600)
	        at androidx.fragment.app.FragmentManagerImpl.moveToState(FragmentManagerImpl.java:881)

android/uhabits-core/src/main/java/org/isoron/uhabits/core/models/HabitList.java
```diff
@@ -228,7 +228,7 @@ public void writeCSV(@NonNull Writer out) throws IOException
             String[] cols = {
```
recheej: todo:  make sure description is also added to this

android/uhabits-core/src/main/java/org/isoron/uhabits/core/test/HabitFixtures.java
```diff
@@ -43,7 +43,7 @@ public Habit createEmptyHabit()
     {
         Habit habit = modelFactory.buildHabit();
         habit.setName("Meditate");
-        habit.setDescription("Did you meditate this morning?");
+        habit.setQuestion("Did you meditate this morning?");
```
recheej: todo: add description to these tests?

recheej:
image

@iSoron what do you think? Maybe text is too strong/dark?

recheej:
@iSoron I noticed you have screenshot tests. How do you generate the screenshots of just the view? Thanks!

android/uhabits-android/src/androidTest/java/org/isoron/uhabits/activities/habits/show/views/NotesCardTest.kt
```diff
@@ -0,0 +1,49 @@
...
+                .inflate(R.layout.show_habit, null)
+                .findViewById<View>(R.id.subtitleCard) as SubtitleCard
+        view.apply {
+            habit = habit
+            refreshData()
+            measureView(this, 800f, 200f)
+        }
+    }
+
+    @Ignore("how do I generate these shots?")
```
recheej: cc @iSoron

iSoron:
> @iSoron I noticed you have screenshot tests. How do you generate the screenshots of just the view? Thanks!

You can write the view test as if the expected screenshot file was already there. If the file doesn't exist yet, it will be generated and the test will fail. The failure message will tell you where the generated file was saved.

recheej:
@iSoron thanks! those tests are fixed now.

android/uhabits-android/src/androidTest/java/org/isoron/uhabits/acceptance/steps/CommonSteps.java
```diff
@@ -19,26 +19,40 @@
...
 
+import androidx.annotation.StringRes;
 import androidx.recyclerview.widget.RecyclerView;
-
-import org.isoron.uhabits.*;
+import androidx.test.espresso.PerformException;
+import androidx.test.espresso.contrib.RecyclerViewActions;
+import androidx.test.uiautomator.By;
+import androidx.test.uiautomator.UiSelector;
+import androidx.test.uiautomator.Until;
```
iSoron: To keep our code style consistent, could you please switch these to wildcard imports?

android/uhabits-core/src/main/java/org/isoron/uhabits/core/models/Habit.java
```diff
@@ -148,9 +148,9 @@ public synchronized String getDescription()
         return data.description;
     }
 
-    public synchronized void setDescription(@NonNull String description)
+    public synchronized void setDescription(@Nullable String description)
     {
-        data.description = description;
+        data.description = description == null ? "" : description;
```
iSoron: For consistency, I think it would be better to keep this method as `setDescription(@NonNull string description)`, and move the null check to whoever is calling this method. In general, I would like to avoid `nulls` being passed around.

android/uhabits-core/src/test/java/org/isoron/uhabits/core/database/migrations/Version22Test.java
```diff
@@ -161,6 +161,7 @@ public void testDisallowNewDuplicateTimestamps() throws Exception
     }
 
     @Test
+    @Ignore("this test is invalid. findAll method queries all columns in the code when the database may be different columns")
     public void testKeepHabitsUnchanged() throws Exception
```
iSoron: Please delete this test instead of just ignoring it.

recheej:
@iSoron thanks! your requests have been addressed.

iSoron:
Thanks for the patch, @recheej!

PR #556, opened 2020-02-13
Add new menu item to sort by score ascending

### LINKED ISSUE #497: Sorting habits by score in reverse order
Hi there,

First of all: a simple but great app! I'm just sorry for all the iOS users who can't use it...

It would be great to offer an option (maybe a check box or a sub menu) to sort the habits by score but in reverse order. This way, the weakest habit would be on top and thus in one's focus on what to work on primarily.

Best regards
macpac59

scrimbly:
Fixes #497 
Adds a new "By lowest score" sort menu item that uses scoreComparator.reverse(). 

Questions (proposals in #497 ):
- What UX do we want? Show two different 'By score' menu items, or reuse the original one?


android/uhabits-core/src/main/java/org/isoron/uhabits/core/models/memory/MemoryHabitList.java
```diff
@@ -151,6 +151,7 @@ public synchronized void setOrder(@NonNull Order order)
         if (order == BY_NAME) return nameComparator;
         if (order == BY_COLOR) return colorComparator;
         if (order == BY_SCORE) return scoreComparator;
+        if (order == BY_SCORE_ASC) return scoreComparator.reversed();
```
scrimbly: I get the following warning in AndroidStudio: 
> Call requires API level 24 (current min is 21): java.util.Comparator#reversed'

Any chance of bumping the minSdkVersion, or shall I just clone the `scoreComparator` and reverse it myself?

scrimbly: I wrote out my own reverse comparator in cc4ad53, so we can keep minSdkVersion=21.
Maybe you'd like a TODO so we can simplify this whenever we do bump minSdkVersion.

iSoron:
Thank you for the pull request! Regarding i18n, unless I am mistaken, it doesn't look like you have modified any strings. If that's correct, then you don't need to do anything. Regarding the user interface, would it be possible to put the icon after "By name", not before, and right-align it?

	+----------------------+
	|  Manually            |
	|  By Name             |
	|  By Color            |
	|  By Score         ↑  |
	|  By Score         ↓  |
    +----------------------+

Also, I would suggest using actual (full) arrows, instead the half arrows you are using currently. The ones you are using currently already mean something on Android (to expand or collapse notifications). Perhaps you could use unicode arrows (↑↓)?

chennemann:
I just read this here and I have some doubts regarding the duplication of the "by score" entry. 
first of all it's confusing that the entry is there twice and after the user figured that out he will wonder why there are no options for the other sorting methods. 
that leads me to my proposal which unfortunately means some more work:

I would suggest to make the name, the color and the score option sortable in both ways (ascending and descending). 
then we could display the arrow indicating the sort direction only after the option that is currently used for sorting. 

+----------------------+
|  Manually            |
|  By Name             |
|  By Color            |
|  By Score         ↑  |
+----------------------+

To make it really easy for the user to learn I would additionally suggest two things:
1. highlight the currently used sort option with an accent color (that way it's transparent for the user which sort is currently active)
2. display a toast after the sort gets changed saying something along the lines of: "Sorted by score. Press again to reverse order."

I hope I could bring my point across. In general this feature can be really useful but we should think about the user experience before just adding another sort option that basically does the same thing as another one that is already present.

scrimbly:
@iSoron You are correct that I no longer have any modified strings. I first prototyped it with "By lowest score" before I figured out how to add the up/down arrow icons. I updated my PR description to reflect only open questions.
I'll see if I can put full arrows on the right edge of the menu items for you.

@chennemann I have similar doubts about the duplicate "by score" entry. I listed a few options in https://github.com/iSoron/uhabits/issues/497#issuecomment-585519642. I also appreciate the symmetry of having ascending/descending orders for each sort option, and that won't be too complicated once we figure out what UX we want.
1. Highlighting/indicating the active sort option makes a lot of sense to me, but I wonder whether we should show the arrow indicating a) the direction the habits are currently sorted or b) the direction they will be sorted when the user clicks the menu item.
2. Displaying a toast would be easy and informative. I'll look into adding that.

chennemann:
@scrimbly Regarding 1: I would show the current sort direction. Reason being that it would be consistent with highlighting the active sort option.

iSoron:
Thank you once again for the pull request, @scrimbly. It has been a while since we heard back from you, so I assume that you are no longer working on this. I went ahead and implemented this feature in 3e99d82, but without an user interface. To toggle between ascending and descending, the user needs to click the the same menu item twice. This can be done for name, color and score.

If either you or anyone else would like to submit a pull request to make the user interface nicer (for example, by showing an arrow), that would be very welcome.

PR #558, opened 2020-02-20
Swipe to dismiss a habit for the day

### LINKED ISSUE #272: Allow user to dismiss the habit for one day
What about adding a button for hiding a task for one day?

scrimbly:
Fixes #272 

- [X] User can swipe a habit
- [X] Swiped habit is tagged as dismissed
- [X] Habit list hides dismissed habits
- [X] "Habit dismissed" popup on swipe (imitates 'Habit created')
- [ ] Persist the dismissal so dismissed habits don't return on app restart
- [ ] Add 'Hide dismissed' filter checkbox to show dismissed habits (default: true)
- [ ] Optional: Add 'dismissed' indication in the ShowHabit activity, and a way to un-dismiss
- [ ] Tests: 
 - isDismissedToday() false after midnight
 - visibility resets at midnight
 - dismissed habit filtered from HabitList (unless 'Hide dismissed' is unchecked)
 - other suggestions?

### DIFF
--- a/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsActivity.kt
+++ b/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsActivity.kt
@@ -52,7 +52,7 @@ class ListHabitsActivity : HabitsActivity() {
 
     override fun onPause() {
         midnightTimer.onPause()
-        screen.onDettached()
+        screen.onDetached()
         adapter.cancelRefresh()
         super.onPause()
     }
--- a/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsScreen.kt
+++ b/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsScreen.kt
@@ -86,7 +86,7 @@ class ListHabitsScreen
         commandRunner.addListener(this)
     }
 
-    fun onDettached() {
+    fun onDetached() {
         commandRunner.removeListener(this)
     }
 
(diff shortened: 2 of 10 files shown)


scrimbly:
I figured out how to update the list to hide the empty gap from the recently dismissed habit!
Updated the PR.

iSoron:
Thank you for the pull request, @scrimbly! A few comments:

1. It seems like dismissing a habit does not modify the database in any way. If you close the app (by pressing the back button) and open it again, would the dismissed habits come back?

2. I think having a "Hide dismissed" option is a good idea. It could be checked by default.

3. This feature seems very related to the ability of "snoozing" a habit in the notifications. Perhaps the two should be integrated in some way? For example, by swiping away a habit, I could be snoozing it until tomorrow. This would cause any notifications to disappear. Similarly, if I snooze a habit in the notifications, then the habit should also vanish from the list (provided that "Hide dismissed/snoozed" is checked). This integration may not be as hard to implement as it seems. Please see [this code](https://github.com/iSoron/uhabits/blob/13f4981066b1fb7a0e0220f5cece88dbaae5c41b/android/uhabits-core/src/main/java/org/isoron/uhabits/core/reminders/ReminderScheduler.java#L155). It would also immediately fix the issue in the first comment (snoozing modifies the SharedPreferences).

scrimbly:
@iSoron Thanks for taking a look!
1. You are correct; dismissing a habit currently does not modify the database, so the dismissed habits come back when the app is restarted. Based on https://github.com/iSoron/uhabits/issues/272#issuecomment-569495625 I saved persistence for later. I'll update the PR description to add it to my TODOs.

2. I'll work on the "Hide dismissed" checkbox (default: true) soon.

3. Snoozing ~= Dismissing.. Interesting. I don't personally use the notifications, but I'll play around with it and see what I can learn.

scrimbly:
@iSoron I see what you mean about the similarity between snoozing a notification and dismissing a habit from today's list, although I'm not yet 100% certain that users would always want the same behavior from both. Let me "snooze" on it. ;-)
Another idea: Maybe swiping right could snooze/dismiss until tomorrow, and swiping left would bring up the SnoozeDelayPicker for finer-grained control of the SnoozeUntil time.

I also wonder if swiping should be allowed when 'Hide dismissed' is unchecked. If so, would swiping snooze the habit (and show an informative toast) and then reappear in its previous place in the list?

NellyWhadsDev:
@scrimbly dismissed habits can be muted in style (grey tint) to solve the last issue you mentioned here

jt007:
Really looking forward to this feature. Any news on this?

iSoron:
Closing due to inactivity. If you are still working on this, please let me know and I will reopen it.

Issue #503, opened 2019-08-06
Some activities change state after orientation change


datalossdetector:
I developed an automatic test case generation tool and it detected these problems in your app.
I would like to know if you consider them as failures and, if not, why. Thanks for your collaboration.

1) When I change the orientation of the device screen, MainActivity changes its state. Below there are two different examples: the screenshots were taken before and after the screen orientation change and they are paired (before, after), (before, after)

--- first case ---

A Dialog disappears after the orientation change.

![2019-07-25 13-27-08_before](https://user-images.githubusercontent.com/53469180/62161979-8f38e100-b317-11e9-824f-4a8258bce972.png)
![2019-07-25 13-27-08_after](https://user-images.githubusercontent.com/53469180/62161980-8f38e100-b317-11e9-998e-3f5e3d3b6156.png)

--- second case ---

The options "PASTE, SELECT ALL" disappear after an orientation change.

![2019-07-25 11-02-38_before](https://user-images.githubusercontent.com/53469180/62161981-8fd17780-b317-11e9-8c73-08f349d42253.png)
![2019-07-25 11-02-38_after](https://user-images.githubusercontent.com/53469180/62161982-8fd17780-b317-11e9-98e9-389a60c46aab.png)

2) ListHabitsActivity, when I change the orientation of the device screen, changes its state. Below there are four different examples: the screenshots were taken before and after the screen orientation change and they are paired (before, after), (before, after)

**--- first case ---**

The options "CUT, COPY, PASTE, ..." disappear from the screen.

![2019-07-25 21-21-55_before](https://user-images.githubusercontent.com/53469180/62161807-42eda100-b317-11e9-9251-1b32dae1f2de.png)
![2019-07-25 21-21-55_after](https://user-images.githubusercontent.com/53469180/62161808-42eda100-b317-11e9-8aec-9e1b76d7c4d9.png)

**--- second case ---**

The dialog disappears from the screen.

![2019-07-25 18-54-58_before](https://user-images.githubusercontent.com/53469180/62161809-42eda100-b317-11e9-92da-cd006ce5e974.png)
![2019-07-25 18-54-58_after](https://user-images.githubusercontent.com/53469180/62161811-43863780-b317-11e9-85f3-e59d59e725ff.png)

**--- third case ---**

The selected item is no longer selected.

![2019-07-25 16-08-51_before](https://user-images.githubusercontent.com/53469180/62161812-43863780-b317-11e9-88ba-06d62e83c6c5.png)
[...]

PR #559, opened 2020-02-22
Prevent orientation change from clearing edit text fields

Bryanx:
Partly solves #503.

Upon creating/editing a habit the edit text fields numerator, denominator and description were being cleared upon orientation change. This issue is now resolved.


iSoron:
Thank you for the pull request! To make sure that this bug does not return in a future version, could you try to write an automated user interface test? You can find an example [here](https://github.com/iSoron/uhabits/blob/13f4981066b1fb7a0e0220f5cece88dbaae5c41b/android/uhabits-android/src/androidTest/java/org/isoron/uhabits/regression/SavedStateTest.kt). Please let me know if you have any questions.

iSoron:
The issue this pull request targets has already been fixed in the `dev` branch after we redesigned the entire form. In any case, thank you for your contribution, @Bryanx!

PR #572, opened 2020-04-18
Add strings.xml for armenian language

sirekanyan:


sirekanyan:
Is it a proper way to add a new language to https://translate.loophabits.org? Armenian language is missing now on that platform.

iSoron:
Thank you for the translations, @sirekanyan. Armenian is now available at translate.loophabits.org. Could you please upload them there?

Issue #564, opened 2020-03-20
Resetting all habits and start from the beginning


p32929:
in issue #61 @sojusnik mentioned about a feature to select multiple habits and reset the selected checkboxes. But @iSoron replied with some drawbacks of the feature & completely respect his decision. But how about a feature to reset all the habits without selecting and starting afresh?

in #61 @iSoron mentioned that, we can easily delete and create a new habits. But deleting a lot of habits and re-entering their names, setting reminders for them and also sometimes some people do change the colors, doing all that over and over again for some habits, that's a lot of work to do. A feature to reset multiple/all tasks would be greatly appreciated. This feature can be implemented in the settings activity/page where we cannot go mistakenly and click on the option by chance.

And if you still don't want to add the feature, I'd still respect your decision. In that case, could you please tell us how we can implement the feature in our forked codebase? Which files we need to change and how? That would be greatly appreciated...

Best wishes

iSoron:
> But how about a feature to reset all the habits without selecting and starting afresh?

Thank you for the suggestion, but I unfortunately don't see this as significantly different from what was proposed #61. My comments in that issue are still valid. I have also not seen significant user demand for this feature.

p32929:
I'm not actually asking YOU to implement the feature if you don't want to. I'm just asking, if you could guide us a little bit about which file to change, where to look for, etc that would be really helpful :)

chennemann:
There is a (now closed) pull request for this feature [here](https://github.com/iSoron/uhabits/pull/592). You might have a look there to see how to solve your particular issue.

p32929:
Thanks :) Really appreciate :)

Issue #61, opened 2016-03-22
Reset habits


sojusnik:
Reseting the progress of habits through multiple selection (like it's already implemented with colors) would be nice to have.

iSoron:
By resetting the progress, do you mean deleting all the checkmarks? That would be a very destructive operation, and not used very often.

sojusnik:
Exactly. Sometimes its more motivating to start over fresh, by deleting all checkmarks of a habit.

iSoron:
I have decided not to implement this feature for the following reasons:
1. As I said, this is a very destructive operation. There is always a possibility that you may press it by mistake on the menu and end up confirming it. We already have one such operation (delete); let's not put even more.
2. There is not a lot of metadata about habits anyway, so "resetting" a habit can be easily accomplished by deleting it and then creating a new one with the same information.
3. It goes against the philosophy of the app. I understand that many other apps are focused on streaks, and that, after failing a streak, you might want to delete everything and start from scratch, since you have already lost all your progress anyway, but Loop is not like that. Every repetition in the past counts towards your score; it doesn't matter if it was one month, or even one year ago. Instead of restarting frequently, the app should encourage you to pick the habit up as soon as you can, and forget about a few missed days (or weeks).

GutHib:
As this feature has been asked for again and again (and again and again) over the past nine years, I would like to comment on the reasons why it hasn't been implemented, because I think that the app has changed so much over the years that it may be time to reconsider.

#3 Concerning the philosophy of the app: As I pointed out in a different thread, to me the prime reason to do this is not failed streaks, but incomplete data. To my knowledge, back in the days the data you could enter was very basic, now you can track pretty much whatever you like and enter any data you want to. Tracking the meters you walked, the calories you ate or the words you wrote is an effort in itself, giving up on tracking isn't the same as giving up on a habit (obviously, you will consume calories whether you track them or not). If you want to track a bunch of very different things at the same time but circumstances kept you from entering data, starting from scratch can be the best (or even only) fix.

#2 If you track a number of things and have everything set up as widgets, creating it all from scratch can be quite an effort. And again, this is because the app has become so versatile.

#1 Whether or not it is destructive depends on how it is implemented. My preferred option would be to set this on a per habit basis, and possibly have a "start tracking" date rather than a downright delete button. This could hardly be used by mistake.

If you still don't want to add it, that's fine - just thought I'd point out that some of the arguments may not be as valid anymore as they used to be. Thanks and keep up the great work!

PR #592, opened 2020-06-15
Reset functionality pr

pkorove:
Hi there,

We have implemented the code for resetting a habit as requested and described in #564  

What the code does is that it deletes all repetitions in a habit. The way to access the feature is by either long pressing one or more habits and tapping on the three dots and then tapping reset under the delete option or by going to the settings in the database tab. The reset option in the settings menu resets all of the users habits. We have also added a confirmation dialogue screen just like when deleting a habit. 

The code was tested with Android Studio through a virtual device and it also includes junit test cases for the class of the reset command.


iSoron:
Thank you for the pull request, @pkorove. It seems well written, but my comments in #61 are unfortunately still valid.

PR #596, opened 2020-06-17
Reset habit strength graph scroll location when switching scale

### LINKED ISSUE #575: Habit strength graphs don't reset when switching scale
When you scroll back the graph I'm day view and then switch to quarter the graph goes blank.

alxmjo:
This PR closes a bug in which the habit strength graph does not correctly reset after switching scale (days, weeks, months, quarters, etc.). It does this by making the following changes:

- Create `ScrollableChart.reset()` to set `x` coordinate of scroller
- Call `reset()` from `doInBackground()` in `ScoreCard` to reset scroll location

Closes #575

### DIFF
--- a/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/views/ScrollableChart.java
+++ b/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/views/ScrollableChart.java
@@ -217,6 +217,13 @@ public abstract class ScrollableChart extends View
         scrollController = new ScrollController() {};
     }
 
+    public void reset()
+    {
+        scroller.setFinalX(0);
+        scroller.computeScrollOffset();
+        updateDataOffset();
+    }
+
     private void updateDataOffset()
     {
         int newDataOffset = scroller.getCurrX() / scrollerBucketSize;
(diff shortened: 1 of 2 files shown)


alxmjo:
From the failed build: 

```
>>> Fetching logcat 
##[error]The process 'android/build.sh' failed with exit code 1
```

I attempted running `build.sh` locally, but it failed there also. I figured it was a bug with the build or test itself, but perhaps I'm mistaken.

iSoron:
Thank you for the pull request, @alxmjo. Could you also fix `HistoryCard`? It's also affected by the same bug.

Regarding the build failure, it seems like we have a flaky test. I will disable that test for now.

alxmjo:
I'm unable to reproduce the bug in `HistoryCard`. Do you want it to just reset its scroll position whenever the scale is changed above **Habit strength**? If so, I'm assuming that adding `chart.reset()` within `HistoryCard`'s `doInBackground()` will do the trick.

alxmjo:
For some reason the app I've been using on my phone doesn't allow changing scale in history card (what I see as **Calendar** in the `dev` branch is `History` on my phone – perhaps this change just hasn't made it into production yet).

Anyway, I tried the simple fix I suggested above without any success (the scroller position doesn't reset for the history card), so I'll have to dig into this a bit more. I'll be off the grid till Sunday but can have another go then.

iSoron:
Merged. Thank you for the pull request! Please feel free to submit another one to fix the other view.

> For some reason the app I've been using on my phone doesn't allow changing scale in history card (what I see as Calendar in the dev branch is History on my phone). Perhaps this change just hasn't made it into production yet.

This is a new feature in version 1.8. You can get it through Google Play by becoming a [beta tester](https://play.google.com/apps/testing/org.isoron.uhabits).

Issue #342, opened 2017-10-05
Allow Checkmark Widget to show current streak instead of current score


Teaminds:
I was very motivated by the idea of "unbroken chain" - that is, the knowledge that I was "doing" the problem without permits has been so many times in a row. For example, "You publish records on Facebook for 24 days" - it literally controls me.
I'm happy to use the application on my Android smartphone, but I'm very uncomfortable that on the screen with widgets-habits (I have a lot of them) I do not see the chain length for every habit (and, ideally, how many days before the break is left - I there are a few habits "once a week" and "4 times in 30 days").

iSoron:
Thank you for the suggestion, @Teaminds. I guess we could allow the user to choose whether the checkmark widget should display the current score or the current streak. In case we show the current streak, it would probably also be a good idea to show the best streak so far. I am not sure how to represent this visually, though. I would like to avoid having simply a pair of numbers, if possible. If anyone has any ideas and want to create a quick mock-up, that would be very welcome.

mrgoodrich:
+1 to this suggestion. Regarding the visualization, as a user I would be fine with numbers for the minimum viable product @iSoron , especially if it was an option in settings. The numbers could appear after the checkmark is clicked.

dkabambe:
+1 from me too. I'd be happy to just have current steak on widget as that's the key at a glance info.

Would be good to have current and longest steak on the main home screen, (maybe only in landscape, on bigger displays, or as optional second line for each habit?)

spitofland:
I think just the current streak would be the most relevant information to the checkbox widget, because it's what you are currently working on. Trying to fit the best streak as well might get tight. Maybe have an option to just include the current streak and another option to also include the best streak.

For display, I think the best option would be to put the number(s) in the middle of the circle where the 'x' and check mark currently sit. If both numbers are displayed, I would present it like a fraction (current/best with a horizontal line between them). To help users understand, you really only need to use a similar notation inside the app. For example, add the fraction (current/best) on the main screen after the habit name, to the "Overview" on the habit page, and/or in the "Best streaks" section of the habit page. You could also tell the user what it will look like in the settings with something like "Show the Current and Best Streaks in the checkbox widget as a fraction (current over best)".

I would always show these numbers for motivation. The color coding change should be enough to signal that the box has been checked, though you will want to check with some color-blind users.

NegatioN:
[...]

PR #499, opened 2019-07-25
Update checkmark widget for numerical support

Tthecreator:
While I'm still waiting for my other pull request #498 to receive feedback, I've started working on making the checkmarkWidget work with numerical habits as discussed in #42.

Work to do:
- [X] Display numerical values in the widget
- [x] Spawn a menu asking for a new numerical value when a user pressed the widget
- [x] Update color of widget based on if today's goal was reached

~~**NOTE** This has been branched from #498. #498 should be improved and merged before this one can.~~

PR #600, opened 2020-06-23
Widget Displaying Current Streak Instead of Checkmark

NegatioN:
I think it would be great if we had a 1x1 widget displaying the current streak, instead of the checkmark. Functionality could otherwise be identical to the checkmark-widget IMO.

I misunderstood this to be linked to #342 , but that request seems to be slightly different.

So before I do any cleanup-work, and details, I was wondering if this is something that would make it into the main branch?

In case it's alright to merge it in, are there any changes you would like in the code? Currently it's pretty much a duplication of the `CheckmarkWidget`, except displaying a different text. I thought about having it as a preference in the widget itself, but there's not really any huge negatives to separate it out in it's own class IMO.

Here's a screenshot of how it looks like compared to the `CheckmarkWidget`
widget


iSoron:
Thank you for the pull request, @NegatioN. The main issue I have with this proposed idea is that, when we introduce numerical habits in Loop 2.0, your widgets will become indistinguishable from #499. I'm okay a togglable 1x1 widget displaying streaks, but I would like a different visual representation, since we are showing a different piece of data.

Below is a concept I came up with a while ago, but I think it could be improved.

LoopHabitTracker

NegatioN:
Thanks for the quick response @iSoron. That's a good point, the widgets might definitely "collide" in a sense.
What do you think about simply having a "headline" of some sort on top to differentiate it?
85467288-4fbc5880-b5ab-11ea-9e63-2d6119740f67

One thing I worry a bit about with the design you've proposed (although it definitely looks nicer from a design perspective), is that it might look very messy if you actually manage to keep your streaks going for a while. Even at 30-40 it seems to get quite crowded there? I guess there could be different dots for different "sizes" or something though, like a single instance showing as 1 dot, but 10 of them showing as a slightly larger dot for example?

Also, although the numbers-widget might be easily confused with a streaks-widget, it doesn't really make sense to have both of them for the same Habit at any point? (i.e. having a streaks-widget for a number-Habit) So that in itself should give *some level of clarity* (still agree it's a challenge though)

NegatioN:
Closing this PR as my wishes seem to stray a bit from the project direction.

Implemented a Green --> Red fade on widgets based on the time remaining until the "chain breaks", and a simple counter instead of the checkmark. Feel free to contact me if there's any obvious way to integrate something similar in the main branch. :slightly_smiling_face:

PR #605, opened 2020-07-17
Add German translations

quelbs:
Adding new German translations using a pull request as the process with crowdin seems to be broken.


quelbs:
It looks like the attribute `formatted="false"` for some translation strings did break the translation sync with https://translate.loophabits.org/. With this pull request I would like to add the German translations for strings added after December 2019 to the end of the file.

iSoron:
Thank you for the pull request, @quelbs. I have updated the translations on Crowdin. Could you please submit them there? Also, please let me know if you would like to be a proofreader there. This will allow you to correct previous translations.

quelbs:
I updated the translations and would be fine to be proofreader at Crowdin.

PR #608, opened 2020-07-20
Only save numerical habit features if the habit is numerical

KristianTashkov:
Currently in the dev branch saving YES_NO habits is crashing because it can't parse the empty string value of `targetInput.text` as double.


iSoron:
Thank you for the pull request!

Issue #88, opened 2016-04-25
Allow user to skip a day without breaking streak


ab014:
A user should be able to skip a day without breaking the streak if he was unable to perform activity due to external factor. For example, the swimming pool was closed due to a public holiday.

andunix:
It should also be possible to always skip habits on some weekdays. If you can execute a habit only at work (clean desk for example). It should be possible to do this habit five days a week and reacht 100% as fast as other habits.

iSoron:
I think the need of having "skipped days" comes from the fact that streaks are a very poor measure of habit strength. In my opinion, they are inherently flawed. There is ample psychological research showing that a few lapses have negligible impact in habit formation. The formula we currently use to compute strengths does not suffer from this problem. A single lapse after a long streak will have almost zero impact on the score. Only if the lapses become frequent enough, the score will suffer.

This is one reason, by the way, why the app focuses so much on the strength, and not streaks. Until we come up with something to fix the streaks somehow, they will be second-class citizens. One idea would be counting how many days in a row you have maintained a score higher than, say, 90%. In this way, a few lapses would not break the chain.

To summarize, I think the need for this feature comes from an inherent flaw on streaks, so I am not very enthusiastic to add it to the app. Feel free to convince me otherwise.

iSoron:
> It should also be possible to always skip habits on some weekdays

Unless I am missing something, [this is already possible](http://loophabits.org/faq.html#q2).

> It should be possible to do this habit five days a week and reacht 100% as fast as other habits

This is troublesome for habits with very long interval. If we really followed that, after a single repetition of a monthly habit, the score would already jump to 100%. Currently, no matter what the frequency of your habit is, you will need the same number of "perfect repetitions" to reach 100%.

ab014:
I agree with what you are saying. Streaks probably are a poor measure of habit, but they are easily understandable. While, in the present state of the app, strength seems ambiguous with no documentation about what it signifies.
Which is maybe why a user would be more interested in not breaking a streak. Also, maintaining a long streak provides with a sense of accomplishment which will probably stay absent until i understand what strength actually signifies.

torkelsson:
I agree with @ab014. 

I find it quite motivating trying to keep as long streak as possible, especially those dark, rainy days when you really just want to stay in, instead of go out and do your 10000 steps. After doing the steps you feel so much better both physically and knowing that you have kept the streak for one more day. 

[...]

Issue #96, opened 2016-05-10
Distinguish between"not checked" and "not entered"


disser:
It would be useful to enter an X for a day deliberately. For habits such as "eat healthy," if you eat ice cream for breakfast you could immediately mark an X rather than remember at the end of the every thing that was eaten.

iSoron:
Thank you for the suggestion. This feature makes sense, but I don't see how to add it while keeping the interface simple. I am open to suggestions.

One workaround for your scenario would be putting a check mark early in the morning and then, as soon as you fail the habit, you remove it. We could also allow check marks to appear automatically on new days.

PunctualEmoticon:
What about pressing and holding on the desired day (like we currently do to add check marks), then dragging down before releasing, to change the check to an X?  This would maintain the same gesture for marking checks, and wouldn't interfere with scrolling because you have to hold in one spot first.

disser:
That sounds great, though I think it would have to be clear that only the selected check mark was being affected, perhaps by dropping a red X directly below the check mark. That would avoid confusion with affecting the entire goal. Also, there would have to be good separation between the regular check action and the X action, like make it an extra long hold?

battaglr:
This is a very good idea!

PunctualEmoticon:
@disser I was imagining that adding a check mark would be the same gesture as it is now, with a hold and release on the desired spot, and making an X would be a hold, drag down, and release on the spot. Upon beginning the drag, the X would start to appear (either with a morph or fade) and it would phase in completely as the user continues to drag down (about half an inch). To cancel, the user just drags back up to where they started, and the X phases back out.

aleksejrs:
I still fail every day to hold on the right spot to add a check mark, so when that happens, I drag up or down a little to avoid opening the habit's page.  Just saying, as a new touchscreen user.

PunctualEmoticon:
@aleksejrs Mm, good point.  But with what I'm proposing, you could just drag up, left, or right and still be fine.

woj-tek:
A couple of suggestions (also related to the linked issue):
* habits could have tick (done), circle/square (pending), cross (not completed) states
* you could mark as done or not done from notification (3rd option to the current 'mark' and 'snooze')
* you could mark item on the list by long-pressing and it could pop-up a list with "done" or "ignored" (magine emoticons pop-up or alternative character keys on keyboards)

[...]

PR #610, opened 2020-07-21
Skip day checkmark

KristianTashkov:
Following the discussion in #88 I created an implementation of skip days (mostly for myself). This implementation might be a bit hacky in some places but I wanted to get the whole thing done and test how it feels. If there is any interest in merging this in dev from @iSoron I can go back and add more tests and rework some of the logic. For this implementation I got working:
1) Toggling in history and in the main list now has 3 states: UNCHECKED -> CHECKED_EXPLICITLY -> SKIPPED_EXPLICITLY instead of two. If this is too obtrusive we might consider adding a preference to allow skip days so this isn't default behavior.
2) For score purposes days that are skipped don't change the score from the previous days and are ignored.
3) For streak purposes they are treated as if they were checked and don't break streaks. I probably need to implement start/end to never start/end from a skipped day so a streak of only skipped days isn't tracked.
3) Skipped days in habit list are shown with a new icon (first thing that looked okay, I'm open to better ideas)
4) Skipped days in calendar are shown similar to implicit checks but with a strikethrough text (again open for better ideas)

Next steps might be:
1) Add better test coverage for the new logic
2) Add skip current day as a button in reminders
3) Clean up code because I haven't written Java in years.

### DIFF
--- a/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/views/HistoryChart.java
+++ b/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/views/HistoryChart.java
@@ -362,6 +362,7 @@ public class HistoryChart extends ScrollableChart
                             GregorianCalendar date,
                             int checkmarkOffset)
     {
+        pSquareFg.setStrikeThruText(false);
         if (checkmarkOffset >= checkmarks.length) pSquareBg.setColor(colors[0]);
         else
         {
@@ -371,13 +372,18 @@ public class HistoryChart extends ScrollableChart
             {
                 pSquareBg.setColor(isNumerical ? textColor : colors[1]);
             }
+            else if (!isNumerical && checkmark == 3) {
+                pSquareFg.setStrikeThruText(true);
+                pSquareBg.setColor(colors[1]);
+            }
             else pSquareBg.setColor(colors[2]);
         }
 
         pSquareFg.setColor(reverseTextColor);
+
         float round = dpToPixels(getContext(), 2);
         canvas.drawRoundRect(location, round, round, pSquareBg);
-        String text = Integer.toString(date.get(Calendar.DAY_OF_MONTH));
+        String text = Integer.toString(date.get(Calendar.DAY_OF_MONTH)) ;
         canvas.drawText(text, location.centerX(),
             location.centerY() + squareTextOffset, pSquareFg);
     }
--- a/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/edit/EditHabitActivity.kt
+++ b/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/edit/EditHabitActivity.kt
@@ -214,9 +214,11 @@ class EditHabitActivity : AppCompatActivity() {
             habit.setReminder(Reminder(reminderHour, reminderMin, reminderDays))
         }
         habit.frequency = Frequency(freqNum, freqDen)
-        habit.targetValue = targetInput.text.toString().toDouble()
-        habit.targetType = Habit.AT_LEAST
-        habit.unit = unitInput.text.trim().toString()
+        if (habitType == Habit.NUMBER_HABIT) {
+            habit.targetValue = targetInput.text.toString().toDouble()
+            habit.targetType = Habit.AT_LEAST
+            habit.unit = unitInput.text.trim().toString()
+        }
         habit.type = habitType
 
         val command = if (habitId >= 0) {
--- a/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/CheckmarkButtonView.kt
+++ b/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/CheckmarkButtonView.kt
@@ -63,7 +63,8 @@ class CheckmarkButtonView(
     fun performToggle() {
         onToggle()
         value = when (value) {
-            CHECKED_EXPLICITLY -> UNCHECKED
+            CHECKED_EXPLICITLY -> SKIPPED_EXPLICITLY
+            SKIPPED_EXPLICITLY -> UNCHECKED
             else -> CHECKED_EXPLICITLY
         }
         performHapticFeedback(HapticFeedbackConstants.LONG_PRESS)
@@ -95,6 +96,9 @@ class CheckmarkButtonView(
     private inner class Drawer {
         private val rect = RectF()
         private val lowContrastColor = sres.getColor(R.attr.lowContrastTextColor)
+        private val mediumContrastTextColor = sres.getColor(R.attr.mediumContrastTextColor)
+
+
 
         private val paint = TextPaint().apply {
             typeface = getFontAwesome()
@@ -106,9 +110,11 @@ class CheckmarkButtonView(
         fun draw(canvas: Canvas) {
             paint.color = when (value) {
                 CHECKED_EXPLICITLY -> color
+                SKIPPED_EXPLICITLY -> mediumContrastTextColor
                 else -> lowContrastColor
             }
             val id = when (value) {
+                SKIPPED_EXPLICITLY -> R.string.fa_skipped
                 UNCHECKED -> R.string.fa_times
                 else -> R.string.fa_check
             }
--- a/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/show/views/OverviewCard.java
+++ b/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/show/views/OverviewCard.java
@@ -160,7 +160,7 @@ public class OverviewCard extends HabitCard
             cache.todayScore = (float) scores.getTodayValue();
             cache.lastMonthScore = (float) scores.getValue(lastMonth);
             cache.lastYearScore = (float) scores.getValue(lastYear);
-            cache.totalCount = habit.getRepetitions().getTotalCount();
+            cache.totalCount = habit.getRepetitions().getTotalSuccessfulCount();
         }
 
         @Override
--- a/android/uhabits-android/src/main/res/values/fontawesome.xml
+++ b/android/uhabits-android/src/main/res/values/fontawesome.xml
@@ -24,6 +24,7 @@
 	<string translatable="false" name="fa_star_half">&#xf089;</string>
 	<string translatable="false" name="fa_star_half_o">&#xf123;</string>
 	<string translatable="false" name="fa_arrow_circle_up">&#xf0aa;</string>
+	<string translatable="false" name="fa_skipped">&#xf101;</string>
 	<string translatable="false" name="fa_check">&#xf00c;</string>
 	<string translatable="false" name="fa_times">&#xf00d;</string>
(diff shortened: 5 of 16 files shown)


KristianTashkov:
I had some time on my hands and went a little bit further on this idea. I saw the discussion in https://github.com/iSoron/uhabits/issues/96 and decided that it might be useful to implement something along those lines as well. At the end this is what I have in this PR:
1) Unless you turn on "Enable advanced options" in preferences, the app behaves exactly like before.
2) If it is turned on, short pressing (or long press if you have turned on short press to mean toggle) on a checkmark in the main screen now opens a popup dialog for you to select an option for the current day:
image
3) "Skip"ing a day pauses the score for that day and acts similarly to an implicit check for streaks, with the difference that streaks can't begin or end on skip days. Visually it looks like a dash in the list.
4) "Fail"ing a day has two states. If it's done on a day that the habit had to be done, it is the same as leaving it unchecked, with the difference of "Show Completed" filters out the habit in the list. This is very convenient if you like the "inbox zero" approach and don't want to keep seeing the habit that you failed the entire day. Visually it looks like a slightly lighter X. If that day you had an implicit check, you marking it as "failed" is ignored in all calculations and it is  treated like before, and only shown visually (slightly brighter check than the normal implicit day).
These are all the states visually (`CHECKED_IMPLICITLY`, `CHECKED_EXPLICITLY`, `FAILED_EXPLICITLY_UNNECESSARY`, `CHECKED_EXPLICITLY`, `FAILED_EXPLICITLY_NECESSARY`, `SKIPPED_EXPLICITLY`, `UNCHECKED`):
image
5) The history card was a little bit harder to distinguish visually but this is what I ended up with, I will be very happy if we figure out something better
image
Basically skips are shown as implicit successes with a line in the middle, fails are shown with a cross and they take background color depending if there was an implicit check or not. Clicking on a day to edit opens the same dialog with options.

TODOs:
1) Better test coverage
2) Notification support for the new functionality: Skip day as a button? Clicking "No" marking an explicit fail?
3) Widget support
4) Prettify the UI, suggestions are welcome

I'm using this branch on my phone and will continue dog fooding it. @iSoron What is your opinion on this approach? The application has it's charm in it's simplicity and I'm sure some users will never use these options which is why I wanted to make sure not to break any default behavior and add it as an "advanced" option. But at the same time following the older threads there are a lot of people that are looking for this sort of behavior (including me) so it seems worthwhile to have in the app.

chennemann:
I really really like this idea and I'm grateful that you took the time to implement it. When I'm back from my vacation I'll try to make some time to test your branch and maybe help polishing so we can integrate it into the main application.

iSoron:
Thank you for the pull request, @KristianTashkov! I will have a look at the code soon. I would be glad to add this feature the app as long as it's off by default, like you suggest. A few comments:

1. For consistency with the notifications and other elements in the UI, I suggest changing done/fail to yes/no.
2. I don't think we need the "delete" option; just yes/no/skip seems sufficient to me. I'm also not sure if the visual distinction between `UNCHECKED` and `FAILED` is necessary. When we implement #96, we can make that distinction more clear. For now, I would suggest making them visually identical to make the interface more simple.
3. The filter behavior you describe seems unexpected to me; I understand hiding skipped habits, but I think all `FAILED_*` states should still be displayed when "Hide completed" is checked. Snoozing a failed habit until the next day is separate issue.
4. I'm concerned about the number of checkmark states we now have. Before this pull request, there were only three (`CHECKED_EXPLICITLY`, `CHECKED_IMPLICITLY` and `UNCHECKED`). Given the two comments above, I think this feature can be implemented with just four states (`CHECKED_EXPLICITLY`, `CHECKED_IMPLICITLY`, `UNCHECKED` and `SKIPPED`). I suggest merging all `FAILED_*` into `UNCHECKED`. Repetitions for boolean habits could now have either `CHECKED_EXPLICITLY` or `SKIPPED` states.
5. Regarding renaming the `Repetition` class, I also think `Entry` would be a better name, but I would prefer if this refactoring was done separately from everything else.

KristianTashkov:
Hey @iSoron, thanks for the comment!
1. I never really liked "fail" as a UI element so moving to YES|NO|SKIP makes sense. I can also change the title to incorporate the question if it's present, similar to a notification.
2. I was also struggling with the purpose of failed and unchecked especially in regards to the UI. In the latest commit (I haven't uploaded updated screenshots yet, because I wanted to also fix the dialog first), there is an actual distinction between "not entered" and "no/failed".
image
This is the latest I have and use on my phone showing all 5 checkbox states for a once every 3 days habit.
3. That makes sense, I can make sure the current completed filter is more in line with what it was previously (if it is checked in some way, hide it, otherwise show it). Then it makes sense to have another filter that is "not entered" vs "resolved" but that might be a separate discussion.
4. This again moves away from #96 and the ability to distinguish not entered vs explicitly failed. Try the latest version with unchecked state being marked as "?" (colored if it's CHECKED_IMPLICITLY), and give your opinion. What you described with the 3 states is just not implementing "explicit fail" as an option, which was the first iteration of this PR.
5. Agree!

This PR is starting to become pretty big with multiple issues being implemented at once. We can try to split it and implement separately if it's easier to discuss and merge:
1) Make implicit checks have 50% opacity, but still colored. This mirrors the history card UI and it's honestly better looking to have completed days all have some color in them.
2) Explicit skip days with the advanced option dialog.
3) Making the distinction between unchecked and explicitly failed. (extra state in advanced option dialog)

KristianTashkov:
Based on @iSoron comments updated the PR:
1) Dialog options now has YES/NO as options and renamed checkbox state not to use the term "FAILED" but instead use "UNCHECKED". Also incorporated the custom habit question into the dialog. It still needs some color in my opinion, especially tying it to the habit color somehow.
image
2) Fixed `Show Completed` filter to filter out both "unchecked" states

KristianTashkov:
Thank you for the detailed answer!

1. To be honest I'm not a frontend developer and this is the best I could come up with in terms of design. If you have a better suggestion I can try to implement it. What don't you like about the latest version of the dialog I have, I've been using it for a month and it feels very fresh and easy to use.
2.  In what way do you mean confusing. In today's app short pressing is only used to show a warning to long press so it's not clashing with any existing behavior (which was my intention). I'm guessing that choice was made because short pressing can be error prone, but with opening a dialog colored in the habit's color mentioning the date, it's very hard to make a mistake.
3. I can somehow agree on the streaks part that skip days shouldn't change anything, and that is why I made that change in a separate commit so it's easy to drop. I still see some value in ranking a week where you did the habit every day (7 day streak) higher than a streak where you did it twice, but skipped the middle 5 days (which would be 2 day streak in my PR, 7 day streak in master). For the score part it feels wrong that "skipped" days would give you a positive boost to your score. The whole idea behind them is to not feel like you are cheating by marking the day checked. If they affect score they are no different than saying you did it that day (except visually).
4. The only reason there are 2 "failed" states is so we can distinguish "checked implicitly" visually. In my last design `?` was used to denote an unknown state and color was showing whether it's an implicit success or you are expected to do the habit today. If you don't have 2 separate fail states you can't show that, and showing an implicit check breaks the idea of "unknown". While I agree filtering by entered can be considered a separate feature that can be implemented in another way, having the "entered" filter is a natural continuation of the idea of unknown/answered. The way I'm using it these days is that it's always turned on, and I treat the list as "pending" items. If I can't do something anymore, I mark it failed and it disappears (e.g don't eat sweets). Everything in the list still requires either me to do a repetition or me to provide an answer whether I succeeded or failed. That way at the end of the day I can mark of everything with success/fail until the list is cleared and then I know there is nothing more to track.

> I don't think users will add skip days that frequently, so I'm okay with it being slightly inconvenient. That being said, I don't quite understand why do you think the double-toggle solution is more inconvenient than the dialog when adding entries retroactively; both solutions require two taps per day.

I disagree on both counts. I have 3 work habits that I mark as skipped every weekend (so twice per week for 3 habits). This week i'm also going on a vacation so I mark work habits as skipped every day. Why I said it's more inconvenient is that today it's two HOLDS not taps, making the process slightly more annoying. Also the dialog opening in the same location makes setting the value with muscle memory very fast. I don't mind the current solution that much, but it's only a crutch solution until we have a fourth option (like fail), and then it becomes not viable.

Merging with master with your changes and my own separate branches clashing is starting to be too much, so I will wait until my other PR is merged/decided upon, before opening this again with unchecked/unknown.

chennemann:
I still haven't had time yet to look into the code of this feature, but I'm closely following the discussion because I'm very interested in it.

I'll try to give a coherent answer to multiple parts of your conversation.

### Checkmark States

> 4. I'm concerned about the number of checkmark states we now have. Before this pull request, there were only three (`CHECKED_EXPLICITLY`, `CHECKED_IMPLICITLY` and `UNCHECKED`). Given the two comments above, I think this feature can be implemented with just four states (`CHECKED_EXPLICITLY`, `CHECKED_IMPLICITLY`, `UNCHECKED` and `SKIPPED`). I suggest merging all `FAILED_*` into `UNCHECKED`. Repetitions for boolean habits could now have either `CHECKED_EXPLICITLY` or `SKIPPED` states.

I do agree that we should be careful with adding new Checkmark states since they make the whole codebase more complicated and the process harder to reason about. However I do think it is very valuable to have a state for each actual state :D

- Unchecked (when nothing is entered)
- Checked (when the habit is completed for the day)
- Skipped (when the habit is skipped for the day)
- Failed (when the habit is failed for the day)

I think a big reason why adding more states is a problem in the first place is the encoding of the source (implicitly/explicitly) into the states which then multiplies the states exponentially.
To solve this I would propose to have an additional boolean flag that indicates whether it is an explicit input or the result of our computation. 

### Dialog UI

>The idea of showing a dialog allowing the user to pick one of multiple options is good, but I think the UI you proposed needs further refinement before making it into the app. For new UI concepts, I suggest making a mock-up first.

I agree with that one. The dialog is functional (and now looks way better than in the beginning) but it does not quite fit into the visual style of the application. However I think it is sufficient for the moment until we have a better fitting alternative.

### Streak/Score Computation

> 4. I reverted the changes to the computation of scores and streaks. In the merged version, "skipped" days are essentially the same as "checked explicitly".

> 3. To be honest I don't fully agree with this interpretation. To me "skip" days are days for which the habit wasn't viable to be either completed or failed, it has no value. As such it should not affect any score or streak in the app, it's something to be ignored.

We have basically two approaches here. One where "skipped days" are basically only a visual feature and the other where skipping a day has actual implications for the score. I do agree with @KristianTashkov here that skipped days should neither affect the score or the streak. Else you could cheat the system by skipping your habit everyday and still reach 100% score.

KristianTashkov:
I agree with everything @chennemann said and having a separation of implicitness and the actual state makes sense a lot.  
The four states that @chennemann mentioned are exactly the four symbols I used: `?`, `✔`, `-` , `X` .  Beside them I used color to denote implicit/explicit success difference. These should be the only values that we ever save in a repetition. With that said the UI has a lot more nuance mostly coming from the fact that some days don't require a repetition, converting unknown/failure for that day into an implicit success. In the future I can see a feature where you can have implicit skipped days (marking a habit as always skipped on weekends if it's a work habit for example), which can also add complications in the UI.

Issue #607, opened 2020-07-20
Re-add different midnight time (3am) for when the "next day" starts


slix:
Starting the day at 3am was added in a75a27ad428e7cacb42ba12c2006b3b688f61b54 . But it seems to have disappeared during a re-write?

I'm a night owl and use the 1x1 habit widgets. I often am still completing habits after midnight, and that makes the habit widgets a headache to use (and somewhat demotivating).

Is https://github.com/iSoron/uhabits/blob/dev/android/uhabits-core/src/main/java/org/isoron/uhabits/core/utils/DateUtils.java#L170 the right place to change?

iSoron:
I reverted those changes because it caused some strange issues with reminders and notifications. This was indeed how I implemented it before, but I no longer think modifying this function is the way to go, because this function is used in other places where the correct timestamp is required; for example, when scheduling reminders. We should probably create a new function for this delayed midnight and go over the source code to decide, in each case, which one should be called. Pull requests are very welcome for this.

PR #621, opened 2020-08-16
Setting custom start of the day

KristianTashkov:
Implements https://github.com/iSoron/uhabits/issues/607.
1) Instead of hard-coding 3 AM like the last time this was implemented, I decided to add it as a setting so everyone can set whatever time they desire to start the new day.
2) Tested it through the app, widgets and notifications and checks go to the appropriate day, UI doesn't move to the next day until you pass the custom mark and notifications arrive as usual so they aren't affected by the offset.
3) It was bugging me that widgets don't automatically refresh when the new day starts, so also implemented that in this PR (even if we don't merge this, we should add this to the main app).

As usual waiting for feedback from @iSoron before I start writing tests for the functionality and will be dog fooding this version in the mean time.

### DIFF
--- a/android/uhabits-android/src/androidTest/java/org/isoron/uhabits/BaseAndroidTest.java
+++ b/android/uhabits-android/src/androidTest/java/org/isoron/uhabits/BaseAndroidTest.java
@@ -88,6 +88,7 @@ public class BaseAndroidTest extends TestCase
         testContext = InstrumentationRegistry.getInstrumentation().getContext();
 
         DateUtils.setFixedLocalTime(FIXED_LOCAL_TIME);
+        DateUtils.setStartDayOffset(0, 0);
         setResolution(2.0f);
         setTheme(R.style.AppBaseTheme);
         setLocale("en", "US");
--- a/android/uhabits-android/src/main/java/org/isoron/uhabits/HabitsApplication.kt
+++ b/android/uhabits-android/src/main/java/org/isoron/uhabits/HabitsApplication.kt
@@ -25,6 +25,7 @@ import org.isoron.androidbase.*
 import org.isoron.uhabits.core.database.*
 import org.isoron.uhabits.core.reminders.*
 import org.isoron.uhabits.core.ui.*
+import org.isoron.uhabits.core.utils.*
 import org.isoron.uhabits.utils.*
 import org.isoron.uhabits.widgets.*
 import java.io.*
@@ -59,6 +60,7 @@ class HabitsApplication : Application() {
             db.renameTo(File(db.absolutePath + ".invalid"))
             DatabaseUtils.initializeDatabase(context)
         }
+        DateUtils.setStartDayOffset(3, 0)
 
         widgetUpdater = component.widgetUpdater
         widgetUpdater.startListening()
--- a/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/views/BarChart.java
+++ b/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/views/BarChart.java
@@ -391,7 +391,7 @@ public class BarChart extends ScrollableChart
     private float getMaxDayWidth()
     {
         float maxDayWidth = 0;
-        GregorianCalendar day = DateUtils.getStartOfTodayCalendar();
+        GregorianCalendar day = DateUtils.getStartOfTodayCalendarWithOffset();
 
         for (int i = 0; i < 28; i++)
         {
@@ -406,7 +406,7 @@ public class BarChart extends ScrollableChart
     private float getMaxMonthWidth()
     {
         float maxMonthWidth = 0;
-        GregorianCalendar day = DateUtils.getStartOfTodayCalendar();
+        GregorianCalendar day = DateUtils.getStartOfTodayCalendarWithOffset();
 
         for (int i = 0; i < 12; i++)
         {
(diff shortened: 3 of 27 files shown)


iSoron:
Thank you for the pull request, @KristianTashkov. Unfortunately, this pull request tries to accomplish too many things at the same time, which makes it very hard for me to review it. Could you please:

1. Revert all class and method renames (for example, from `MidnightTimer` to `DayStartTimer` and `atMidnight` to `atDayStart`). This refactoring can be submitted as a separate pull request after this one is merged.
2. Revert the automatic update at midnight. This can also be submitted as a separate pull request.
3. Revert the changes to the settings screen. This can also be submitted as a separate pull request.
4. I noticed that you replaced most `getToday()` calls by `getToday(true)`. Instead of doing this, please modify `getToday()` to do whatever `getToday(true)` is doing, then create another method `getTodayWithoutOffset()` that does whatever `getToday(false)` does. In general, please try to avoid boolean arguments; they are usually considered a code smell.

KristianTashkov:
@iSoron Thanks for the feedback!
Reverted anything not required to make a fixed custom start of the day work (hard coded at 3 AM like before).
I opted to name the new methods `getTodayWithOffset` instead of making current ones `getTodayWIthoutOffset`. It feels like this is less confusing if someone is trying to work on generic time functions through `DateUtil`. Technically the opposite argument is true that someone might introduce a new function that doesn't account for offset if working on habit logic, so I'm not sure which one is better. If you disagree with the current variant I can global replace the names easily.

iSoron:
> I opted to name the new methods getTodayWithOffset instead of making current ones getTodayWIthoutOffset.

Thanks, that's fine. I will review it soon.

iSoron:
I am getting an exception after the following steps:
1. Create a daily habit with a reminder at any time of the day
2. Set your system time to 1:30 AM to a day in the future; make sure that the notification appears
3. Click "YES" on the notification

The exception is below:
```
    java.lang.IllegalArgumentException: timestamp is not valid
        at org.isoron.uhabits.intents.IntentParser.parseTimestamp(IntentParser.kt:56)
        at org.isoron.uhabits.intents.IntentParser.parseCheckmarkIntent(IntentParser.kt:36)
        at org.isoron.uhabits.receivers.WidgetReceiver.onReceive(WidgetReceiver.java:77)
        at android.app.ActivityThread.handleReceiver(ActivityThread.java:3379)
        at android.app.ActivityThread.access$1200(ActivityThread.java:199)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1661)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6669)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
```

If we remove the check that causes the exception, the app does not crash, but the checkmark is added to the incorrect date. Before fixing this crash, could you try to write an automated test to detect it?

KristianTashkov:
Hey @iSoron did you change anything before getting this exception, because I can't reproduce it.
I hit this problem while developing the branch but have since accounted for it. It happens when you apply two `WithOffset` functions to the same timestamp, essentially moving it back two days. For notifications there are two possible places where that can happen and I left one of them intentionally to use the regular without offset function to account for that:
1) ReminderScheduler.java:124 where we do use with offset function to set the correct timestamp for the notification
2) CheckmarkIntentData:53 where we don't use with offset because the timestamp already shows the correct day it should be used for.

To get the behavior you explained it means the second location also applies the offset and it shouldn't.

iSoron:
@KristianTashkov Never mind, it seems like I accidentally showed the notification, then set the clock to a time before the notification was supposed to appear. This is an issue, but it's not related to your patch. The PR looks fine now, but I would like to test it a bit more around changes in daylight saving times, just to make sure nothing weird happens. I will keep you updated.

iSoron:
Merged! Thanks again, @KristianTashkov. Regarding the changes that I asked you to revert:

1. Please feel free to submit a PR to rename `MidnightTimer` to `DayStartTimer`
2. Regarding the settings, I don't think that level of granularity is necessary. I would prefer a simple boolean setting "delay start of the day" which sets the offset to something reasonable, such as 3am.

KristianTashkov:
2) `Reasonable` is subjective. if we are going to show a setting at all, might  as well let people set it to whatever they want, the code already supports it. If you are interested in that PR I can open it, otherwise I'm going to focus on some other features.

Mrnofish:
>    2. `Reasonable` is subjective. if we are going to show a setting at all, might  as well let people set it to whatever they want, the code already supports it. If you are interested in that PR I can open it, otherwise I'm going to focus on some other features.

Definitely.

Yup, I'd say e.g. people who work at night may have a different idea of what is reasonable compared to most.

kmantel:
> > 2. `Reasonable` is subjective. if we are going to show a setting at all, might  as well let people set it to whatever they want, the code already supports it. If you are interested in that PR I can open it, otherwise I'm going to focus on some other features.
> 
> Definitely.
> 
> I'd say e.g. people who work at night may have a different idea of what is reasonable compared to most.

Strongly agree, considering I'm posting this at about 4:30 AM. Some other day-based software (the example I'm aware of is Anki https://github.com/ankitects/anki) have the option to reset a day at any hour, which lets me use them effectively. 


A cursory glance at the code makes it seem like it wouldn't require major changes to support *any* time
https://github.com/iSoron/uhabits/blob/2fc6c67432fcf3390271bbf32694b6e0f1d23cee/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/utils/DateUtils.kt#L37-L38

and that it may only take an interface/preference change
https://github.com/iSoron/uhabits/blob/6af2edb690cac9ba1a54356fa5030c9365dc0db5/uhabits-android/src/main/java/org/isoron/uhabits/HabitsApplication.kt#L73-L77

I'm not familiar enough with Android development to know, but is that correct?

Issue #526, opened 2019-11-25
Add a 2×1 widget showing time elapsed since habit last ticked


mmazur:
Some habits are sparse, but still regular. I can quickly inspect just one of them in a 2×2 calendar widget, while I could fit two at 2×1 in the same visual space.

(When tapped the widget would act just like the 1×1 and tick the habit for the day.)

PR #626, opened 2020-08-25
Timed Checkmark Widget

aglebionek:
Based on issue #526, this widget inherits from the Checkmark Widget and, instead of displaying an 'X' if the habit is unchecked, it displays the number of days since the last time the habit was ticked.
<h3>Design decisions</h3>
The main purpose of this widget is to display the number of days since the habit was last ticked, mostly to easily keep track of sparse habits, so here is how I decided to handle the various checkmark states, to achieve that and to remain consistent with the original Checkmark Widget (CW): </br>
1. If checked explicitly, it displays the checkmark and the bg color changes to the habit color, same way the CW does </br>
2. If checked implicitly, meaning we still fall within set frequency of the habit, the bg color is grayed and a checkmark is displayed, same way CW does, but, in addition, the number of days since last ticked is showed next to the checkmark </br>
3. If skipped it works the same way as CW </br>
4. If unchecked/by default it only displays the number of days since the habit was last ticked </br>

gif


mmazur:
To elaborate on a use case – this is awesome for infrequent habits.

For example – I want to do X every ~month. In practice this means that I'll want to do X somewhere between 25 and 35 days since last occurrence. (A typical monthly habit is highly unlikely to target *exactly* every, say, 31 days; you don't time your daily habits down to the hour and you wouldn't time your monthly ones down to the day either.)

This widget lets me do just that, since I'll notice at some point that I'm past the 25 day mark and will keep it in the back of my mind that I need to do X soonish. Base checkmark widget is not really usable in this scenario, since it'll just at some point indicate that I'm past my goal and I won't even know how much past. (One could use the calendar widget for this, but that takes 4× the screen space and requires one to do head math to figure out how many days have gone by.)

I'm not a fan of using ✓ and × next to the day counter, but the only alternative I can come up with is using colors, which would then be inconsistent with other widget (and potentially problematic considering habits have colors of their own). So imho this implementation is good enough.

(Ideally uhabits, would have built–in support of "every X to Y days", in which case this widget could be extended to color code the number of days, e.g. gray if it's <25days; green for 25–30; yellow 30–35; red 35+ (habit missed). But that's a separate feature and a separate discussion.)

iSoron:
Thank you for the pull request, @aglebionek, and welcome to the project. Unfortunately, as I mentioned in #526, I don't see enough user demand for this feature, so I am not planning to introduce it. Each new feature makes our code base more complex, and we will have to carry this complexity around for many years to come.

PR #627, opened 2020-08-27
Numerical habits: update notifications and detail screen

namnl:
1. Update detail screen -> Allow user to change values in calendar chart
2. Update notifications -> Replace Yes/No buttons by "Enter"


iSoron:
Thank you for the pull request, @namnl, and welcome to the project. I see quite a few changes in your PR that are only modifying the source code formatting (for example, including or removing white spaces). Before I review it, could you please revert all these cosmetic changes?

namnl:
Thank you @iSoron for the comment. I have removed all format changes in the committed code.

iSoron:
Hi @namnl, one issue I am getting on Android 9 is that the picker dialog appears under the notification drawer. See the recording below. Another issue is that the button doesn't work from the lock screen (although this also affects the "later" button). This [stack overflow question](https://stackoverflow.com/questions/18261969/clicking-android-notification-actions-does-not-close-notification-drawer?answertab=votes#tab-top) may help solving the first issue, and [this one](https://stackoverflow.com/questions/34658292/how-to-ask-user-to-unlock-device-on-clicking-notification-action-in-android) may help with the second issue.

namnl:
Hi @iSoron,
Thank you for your review and support! I have updated the commit.

Regarding the difference between "Enter" and "Later" action from the Lock screen (as below recording):
I found that in SnoozeDelayPickerActivity.java, it dismisses dialog from onPause(). 
Is there any issue if I remove this code as we can see that SnoozeDelayPickerActivity is dismissed in onPause() but NumericalCheckmarkWidgetActivity is not?

`protected void onPause()
 {
     if (dialog != null) dialog.dismiss();
     super.onPause();
 }`

LockScreenAction

android/uhabits-android/src/main/java/org/isoron/uhabits/receivers/WidgetReceiver.java
```diff
@@ -105,6 +104,9 @@ public void onReceive(final Context context, Intent intent)
                             data.getTimestamp());
                     break;
                 case ACTION_SET_NUMERICAL_VALUE:
+                    // close Notification drawer
+                    context.sendBroadcast(new Intent(Intent.ACTION_CLOSE_SYSTEM_DIALOGS));
```
KristianTashkov, line 108: I got to the same solution when trying to account for this problem in a different PR. One thing to note is that entering multiple habits with every one closing the notification drawer becomes a bit annoying. I don't have a better solution to this than somehow inlining the answer instead of showing the dialog. I know that some chat clients open a text view inside the notification when you click "Reply", so maybe we can try a similar approach?

android/uhabits-android/build.gradle
```diff
@@ -87,7 +87,7 @@ dependencies {
     implementation 'androidx.legacy:legacy-support-v4:1.0.0'
     implementation "com.github.paolorotolo:appintro:3.4.0"
     implementation "com.google.dagger:dagger:$DAGGER_VERSION"
-    implementation "com.jakewharton:butterknife:8.6.1-SNAPSHOT"
+    implementation "com.jakewharton:butterknife:10.0.0"
```
iSoron, line 90: Please revert these changes, they are not related to the functionality implemented in this PR.

android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/show/ShowHabitScreen.java
```diff
@@ -26,9 +26,9 @@
 import org.isoron.androidbase.activities.*;
 import org.isoron.uhabits.*;
 import org.isoron.uhabits.activities.common.dialogs.*;
-import org.isoron.uhabits.activities.habits.edit.*;
 import org.isoron.uhabits.core.models.*;
 import org.isoron.uhabits.core.ui.callbacks.*;
+import org.isoron.uhabits.core.ui.screens.habits.list.ListHabitsBehavior;
```
iSoron, line 31: Please use wildcard imports.

android/uhabits-android/src/main/java/org/isoron/uhabits/notifications/SnoozeDelayPickerActivity.java
```diff
@@ -46,6 +46,14 @@ protected void onCreate(@Nullable Bundle bundle)
         habit = appComponent.getHabitList().getById(parseId(getIntent().getData()));
         if (habit == null) finish();
 
+        // Unlock device on clicking notification action
```
iSoron, line 49: Please avoid single line comments explaining what do the next few lines of code do (unless you are in the middle of a complicated algorithm). If you feel that the comment is important enough, please extract the lines of code to a new method and give a descriptive name to the method. In this particular case, we could move it into `unlockScreen()`.

android/uhabits-android/src/main/java/org/isoron/uhabits/receivers/WidgetReceiver.java
```diff
@@ -67,7 +67,6 @@ public void onReceive(final Context context, Intent intent)
 
         IntentParser parser = app.getComponent().getIntentParser();
         WidgetBehavior controller = component.getWidgetController();
-        Preferences prefs = app.getComponent().getPreferences();
```
iSoron, line 70: Please revert this, unless it is required. Refactoring is welcome, but please submit it as a separate PR.

android/uhabits-android/src/main/java/org/isoron/uhabits/receivers/WidgetReceiver.java
```diff
@@ -105,6 +104,9 @@ public void onReceive(final Context context, Intent intent)
                             data.getTimestamp());
                     break;
                 case ACTION_SET_NUMERICAL_VALUE:
+                    // close Notification drawer
```
iSoron, line 107: Same comment above about inline comments. In this case, please either remove the comment or move the line into a `closeNotificationDrawer()` method.

android/uhabits-android/src/main/java/org/isoron/uhabits/receivers/WidgetReceiver.java
iSoron: The Android documentation refers to this as [direct replies](https://developer.android.com/training/notify-user/build-notification#reply-action). It may actually work well. This can be a separate PR.

android/uhabits-android/src/main/java/org/isoron/uhabits/widgets/activities/NumericalCheckmarkWidgetActivity.kt
```diff
@@ -51,6 +51,14 @@ class NumericalCheckmarkWidgetActivity : Activity(), ListHabitsBehavior.NumberPi
                                   component.notificationTray)
         widgetUpdater = component.widgetUpdater
         showNumberSelector(this)
+
+        // Unlock device on clicking notification action
```
iSoron, line 55: I see that this is a repetition of the code above. Please extract them both to a common method.

android/uhabits-core/src/main/java/org/isoron/uhabits/core/ui/screens/habits/show/ShowHabitBehavior.java
```diff
@@ -23,6 +23,7 @@
 
 import org.isoron.uhabits.core.commands.*;
 import org.isoron.uhabits.core.models.*;
+import org.isoron.uhabits.core.ui.screens.habits.list.ListHabitsBehavior;
```
iSoron, line 26: Please use wildcard imports.

android/uhabits-core/src/main/java/org/isoron/uhabits/core/ui/screens/habits/show/ShowHabitBehavior.java
```diff
@@ -58,12 +59,29 @@ public void onEditHistory()
...
+            CheckmarkList checkmarks = habit.getCheckmarks();
+            double oldValue = checkmarks.getValues(timestamp, timestamp)[0];
+
+            screen.showNumberPicker(oldValue / 1000, habit.getUnit(), newValue ->
+            {
+                newValue = Math.round(newValue * 1000);
+                commandRunner.execute(
+                        new CreateRepetitionCommand(habitList, habit, timestamp, (int) newValue),
+                        habit.getId());
+            });
```
iSoron, line 72: This is a repetition of the code in `ListHabitBehavior`. Please create a common method.

iSoron (changes requested):
Thank you for the revision, @namnl. I had only some minor comments. After these are addressed, we can merge it.

android/uhabits-core/src/main/java/org/isoron/uhabits/core/ui/screens/habits/show/ShowHabitBehavior.java
namnl: Hi @iSoron, could you share a suggestion for this change?

iSoron:
Thank you for the changes, @namnl. I see now that the last change is a bit more tricky than it looks like, so I went ahead and merged your PR.

namnl:
Thank you @iSoron

PR #629, opened 2020-09-01
Advanced options dialog with explicit uncheck

KristianTashkov:
Reintroduce advanced options dialog from https://github.com/iSoron/uhabits/pull/610 to implement #96 
@iSoron removed the extra checkmark that was bothering you even though I still don't agree it should be removed, but let's start slow. Also removed the extra filter.
I still kept the reworked streaks building logic since the last one relies on `value > 0` always being success which isn't true and still keeping it while going around that makes it unreadable. This is the same type of logic but with a more clear iteration and logic behind when streaks start and end.
Also removed all the extra SKIPPED logic even though I don't agree with the current variant in `dev` but I can introduce that in another PR/discussion.
I kept the dual behavior with the advanced options on click and toggle on long click because I wasn't sure what you wanted to do last time. I'm okay with only having one of them active at a time, so just tell me if you want new behavior to completely replace toggle and I can change it. I've let them both stay for now for simplicity.
Also we should probably rename the old `UNCHECK` to `UNKNOWN`. I tried to do it @iSoron way of having this the new unchecked, and the new state being `UNKNOWN` but it requires a lot bigger refactoring with tests and current code so I reverted it.

I tried to go down the future numeric notification route with `enter` as a single button. It has a problem that the input dialog is still behind the notification tray, so I close it here, but it makes entering multiple values annoying. @iSoron did you have a better plan for numeric's enter we can do here?
cc: @chennemann


iSoron:
Sorry, but it's not clear to me what is this pull request trying to achieve. It seems like you are trying to tackle #96 while simultaneously making significant changes to the GUI. The PR is also trying to introduce changes to the way streaks are calculated, among other things. Please resubmit this as multiple, much smaller PRs.

For implementing #96, I suggest starting with a PR that only modifies the internal representation of checkmarks but has no visible effects to the user interface. After we settle on a good internal representation, we can move on to adapting the user interface. Also, please don't modify the computation of streaks and scores.

KristianTashkov:
> Also, please don't modify the computation of streaks and scores.

I am not changing any logic but incorporating the new state (explicit check). In the streaks code it heavily relies on the fact that `value > 0` means success (which to be honest is a horrible practice for a variable that is essentially enum to use `>` or `<`). I'm refactoring the logic while keeping it the same as not to add hugely ugly code to the current ifs to account for the new state. This PR was based on your initial skipped implementation so it's outdated since you changed it so it may look different.

> It seems like you are trying to tackle #96 while simultaneously making significant changes to the GUI.

If I remove the GUI the other option would be to add the new explicit fail state to the toggle logic which becomes too much. Also I was trying to hide everything behind a preference settings not to change the current behavior. Since #96 is about making a different between unknown and unchecked, this PR does add a new unknown icon the the UI.

>  I suggest starting with a PR that only modifies the internal representation of checkmarks but has no visible effects to the user interface. After we settle on a good internal representation, we can move on to adapting the user interface. 

This will be the third time I rewrite this logic and there is no point in starting again until we decide how we want to handle #96 in the backend. Can you explain what is the minimum PR you want to have #96 implemented? A couple of discussion points to start us off:
1) What would the initial UI be for the unknown state? How do you go from unknown to unchecked, do we add a 4th toggle option?
2) Adding a new state will have a number above 0 in the Checkmark states enum. This means score and streak logic have to be refactored because they assume `value > 0` means success.
3) Do we activate it for everyone or try to hide it behind a UI preference like this PR was trying?
4) How do we want to handle the new state of explicit unchecked? We have a couple of options the way I see it:
   *  Rename the current `UNCHECKED` to `UNKNOWN` and add a new state that is `UNCHECKED`.
   * Keep the current `UNCHEKED` as the new possible state, and add a new `UNKNOWN` state from which checkmarks state default from. Keep in mind this means changes in a lot of places to account for that difference (that 0 is no longer the default state).

On a a related note, one of the changes that was added here (to be honest forgot to revert for now since the first iteration) was making `IMPLICIT_CHECKED` show with 50% alpha color, instead of being dark. I have seen it requested in an issue with regards to widgets, but in general makes sense that every time we have any type of success (whether it's a skip or implicit), the UI should show color to indicate non-failure. @iSoron Are you interested in a separate change like that?

iSoron:
In a recent commit, I have renamed our current checkmark states to `NO`, `YES_AUTO`, `YES_MANUAL` and `SKIP`, so I will use this new terminology from now on. Answering the last question first:

4. I suggest renaming our current `NO` state to `NO_AUTO` and adding a new state, `NO_MANUAL`.  When this feature is done, the user interface can display `NO_AUTO` as either a cross (if the user doesn't care about #96) or as a question mark (if they do).

Answering the other questions:

1. I would not like to see any user-visible changes at this point. Let's add it to the backend first, but make it completely imperceptible in the frontend. That is, `NO_AUTO` should still be rendered as `NO`. When the user toggles checkmarks, it goes `NO_AUTO`, `YES_MANUAL`, `SKIP` then `NO_AUTO` again.
2. We can still keep the convention that `> 0` means success and `<= 0` means failure if we make `NO_MANUAL=-1`. I am also fine with removing `value > 0` comparisons altogether, but this should be done as a separate PR.
3. At this point, let's make it disabled for everyone in the frontend, and let's not even include an option to enable it. This can come later too.

chennemann:
https://github.com/iSoron/uhabits/pull/610#issuecomment-679233117

@iSoron in this comment I proposed the addition of a new flag to indicate whether a check is manual or automatic. I think this would be a good thing to keep the number of states down. It kinda feels like a bad practice to encode the source of the State into it. Check the comment for more details. What do you think about that?

iSoron:
@KristianTashkov wrote:
> On a a related note, one of the changes that was added here (to be honest forgot to revert for now since the first iteration) was making IMPLICIT_CHECKED show with 50% alpha color, instead of being dark. I have seen it requested in an issue with regards to widgets, but in general makes sense that every time we have any type of success (whether it's a skip or implicit), the UI should show color to indicate non-failure. @iSoron Are you interested in a separate change like that?

For widgets, I think that's a good idea. We could make the background color of the checkmark widget be consistent with the square color in the history chart. For the checkmarks on the main screen, I am less certain. I experimented with it in the past, but fading the foreground color makes the component look disabled. For some colors (yellow, for example), the contrast becomes quite low.

Issue #270, opened 2017-03-29
Variable habit frequency


sm-g:
Some habits have changeable repeat interval, i.e. watering of plants should be done more often in summer than in winter. After changing this parameter, app apply new rule to all history (and percentage rise down, like we missed some at past, but we did not). 

Would be cool be remember (and allow edit) repetition settings change in time.

iSoron:
Hi @sm-g, thank you for the suggestion. While it makes sense for the app to remember changes to the habit frequency, I would expect such a feature to complicate the interface and cause quite a bit of confusion. If anyone thinks this can be done in an intuitive way, please feel free to implement a prototype, and I will consider it.

fin0:
What do you think would create the most part of the confusion? The interface to manage frequencies for a list of time periods for a certain habit?
Or do you think the problem would arise more in the evaluation, e.g. the graph and everything else?

vramazani:
@iSoron
a suggestion:
- a button next to the frequency setting inside editing pop-up, which would open a simple history of frequency changes with timestamps [could be extended to a recurrent scheme, e.g. seasonal]
- Using that history to **visualize** the variable threshold by drawing approximate dotted lines over the _bar chart_ !

jklingen:
> Would be cool be remember (and allow edit) repetition settings

I might be wrong, but from the UI perspective, I think most of the complexity of this request is in the words "allow edit". So here's my suggestion for an MVP (which would be absolutely sufficient for me):

_Given the user has an existing habit,
when they change its frequency,
then display a confirmation dialog allowing to choose whether previous scores and streaks shall be re-calculated._

My use case: my 2x per week habit works out fine, so I decided to increase to 3x. It was a bit surprising to me to see all my previous scores drop, and many streaks to break. I expected the previous values remain frozen, with my score dropping just as of today or tomorrow.
However, I am pretty sure some people prefer it that way, especially in an opposite scenario (e.g. when switching from 3x to 2x per week), so I think it would be nice to let the user decide.

Issue #124, opened 2016-06-10
History plot (calendar) confusing filling


rsuhada:
First of all thank you for the amazing app! I've tried many and your app is the best for me! Amazing work!

For me the most important feature is the calendar view - I need to see 1-3 months in the past so that I'm really pushed to not break my habbit. This works almost great - I have just one problem:

I have a habbit that I want to do 3x a week - but every week it is 3 different days... The visualization somehow tries to "guess" the schedule, but since it is irregular I end up with a confusing plot:

![screenshot_2016-06-10-13-52-44](https://cloud.githubusercontent.com/assets/818539/15963660/00cb582e-2f14-11e6-88ab-3feaaf5cb9ce.png)

My proposal would be:
1. simple: add an option to disable the automatic filling of days (so that only days where the user does tick the habit are colored, nothing else).
2. bit harder: add "full week" coloring: simply for each **finished** week in the past, check if the target was filled (e.g. in my case 3x the given week) and if yes, fill all the unfilled days in that week (signifying that the week was successful).

Thank you very much. I can say that the calendar widget occupies 1/2 of my home screen :) - it is a strong motivation and a great product!

iSoron:
Glad you like the app. Could you tell me what did you expect to see, instead of what is currently shown? In your calendar, for example, what days do you think should have been highlighted, but were not?

rsuhada:
In the option 1: I'd see only day I ticked - so only the bright green fields, no dark green fields.

In this option the image would look exactly the same as the screenshot, but with no dark green fields at all.

In the option 2: if the week was successful then: bright green for ticked day, dark green for all the others
Example of week starting 3. April: 4, 6, 8 = bright green, 3., 5., 7., 9. = dark green 
(this dark green highlighting could happen on 10. April - i.e. after the week is finished and it is checked that I indeed checked my target of 3).

In this option the plot would have all green (some dark and some bright) fields during weeks Apr. 3/10/17/24 - these were all successful weeks, i.e. these 4 weeks should all look like the week 17. Apr looks now (i.e. all days green = successful week). 

For the rest of the weeks there would be only bright fields where I checked in (but not on the other days).

The second option is more satisfying (you'd see long stretches of green days if you are filling your target), but if it is too complicated, option 1 would work too.

aleksejrs:
My understanding of the problem:

[...]

Issue #37, opened 2016-03-16
Add option to only count as streak checked days


AnderRasoVazquez:
If we have a habit to be checked 1 time for X days, and we check that habit one time only, we will see on the streak graph that we will have an X days streak for that habit. I think it will be more accurate to count as streak only the checked days.

For example: we have a habit to be checked 1 time for 7 days. I check that habit one time. I would expect a 1 day streak not a 7 day streak.

What do you think?

iSoron:
The idea for the current way of counting is that it is consistent across all habits. If you see the number "14", you know that you have been doing the habit perfectly for two weeks; if you see "365", you have been doing it for a year, and so on, no matter what is the frequency of the habit. By counting repetitions, the length of time becomes unclear for habits with complex schedule.

But I am well aware that not everyone likes it this way, and it would be quite easy to make this more flexible, so I guess we could add an option on the settings menu to decide this.

AnderRasoVazquez:
That would be great.

Weygert:
Most of my habits are on a weekly basis and considering that I am looking at the success of doing them on a much wider timeframe, imho the counting of days with its quite high numbers are a bit confusing. I.e. I also like the idea of having the choice how to count it for each habit separately.

iSoron:
Since the release of the app, there has very little user demand for this feature. Implementing it would make the interface inconsistent and hard to explain (sometimes the number means something, and sometimes it means something else). It would also add complexity to the app, both for the users and developers. For these reasons, I no longer think we should implement it.

PR #630, opened 2020-09-01
Make skips freeze score

KristianTashkov:
The way skip days are currently implemented in the app have a couple of problems. The biggest one of them is that they are no different than a check for the current day, and only distinct visually, making them a visual gimmick.
The desire for skip days comes from situations where a habit can neither be completed or failed, it's just not valid for a given day. In those days the person can add a check not to break score/streak, but then he is getting rewarded for doing nothing and it feels like cheating. Another thing you can do it is let them be unchecked, but then you are being penalized for things outside of your control (sick days, inability because of location, etc.). The way the app currently implements them is the same as adding a check (which feels like cheating and rewards you for doing nothing), but only visually saves that something happened that day.
A couple of concrete examples why the current behavior doesn't work:
### Streaks
1) For a daily habit that I skipped Mon-Wed, did it on Thu, and skipped Friday-Sunday, the app will tell me I have a streak of 7 days. In reality I just did it a single day so my streak should be 1. (**Beginning and ending a streak with skipped days doesn't make sense**)
2) If in January I did a daily habit 7 days in a row, that is a streak of 7. If then in February I do the habit once on monday, once on sunday and skip everything in between, the app will also reward me a streak of 7, even though the one in february I barely did the habit. They shouldn't be equal. (**Skipped days shouldn't be counted as successful days in streak**).
3) Something both implementations get right is that if you skip a couple of days between successful repetitions, this shouldn't end your streak. (**Failing only occurs on a day when that was possible, not skipped and not implicitly checked day with missing repetition**)
### Score
1) Today in the app skipping 67 days in a row gives you 100% score as if you mastered the habit even though you didn't do a single repetition. (**Skipping days shouldn't be rewarding, it should be neutral**)
2) If I have been doing a daily habit successfully for a month (e.g. workout every day) and then have an injury that makes it impossible to continue doing the habit for another month. You shouldn't get rewarded for this month of injury when you weren't working out, but you shouldn't also get penalized for it by losing all your progress. In terms of that habit your score should be frozen until you can do the habit again and you can keep tracking it. The same applies for a work habit that is invalid on weekends, you shouldn't get rewarded or penalized for skipping it during the weekend or during a vacation. (**Skipping days shouldn't penalize you because the habit was not possible to complete**).

This PR implements fixes for those problems:
1) Skip days don't break streaks but are also excluded from the number of successful days you were in a streak to give more days when doing the habit, instead of just skipping it.
2) Skip days don't change the score for the current day and just freeze the score from the previous one.

### DIFF
--- a/android/uhabits-core/src/main/java/org/isoron/uhabits/core/models/ScoreList.java
+++ b/android/uhabits-core/src/main/java/org/isoron/uhabits/core/models/ScoreList.java
@@ -286,7 +286,7 @@ public abstract class ScoreList implements Iterable<Score>
                 double percentageCompleted = Math.min(1, rollingSum / 1000 / habit.getTargetValue());
                 previousValue = Score.compute(1.0, previousValue, percentageCompleted);
             }
-            else
+            else if (checkmarkValues[offset] != Checkmark.SKIP)
             {
                 double value = Math.min(1, checkmarkValues[offset]);
                 previousValue = Score.compute(freq, previousValue, value);
--- a/android/uhabits-core/src/test/java/org/isoron/uhabits/core/models/ScoreListTest.java
+++ b/android/uhabits-core/src/test/java/org/isoron/uhabits/core/models/ScoreListTest.java
@@ -126,6 +126,49 @@ public class ScoreListTest extends BaseUnitTest
     }
 
     @Test
+    public void test_getValueWithSkip()
+    {
+        toggleRepetitions(0, 20);
+        addSkip(5);
+        addSkip(10);
+        addSkip(11);
+
+        double expectedValues[] = {
+                0.596033,
+                0.573910,
+                0.550574,
+                0.525961,
+                0.500000,
+                0.472617,
+                0.472617,
+                0.443734,
+                0.413270,
+                0.381137,
+                0.347244,
+                0.347244,
+                0.347244,
+                0.311495,
+                0.273788,
+                0.234017,
+                0.192067,
+                0.147820,
+                0.101149,
+                0.051922,
+                0.000000,
+                0.000000,
+                0.000000
+        };
+
+        ScoreList scores = habit.getScores();
+        Timestamp current = DateUtils.getToday();
+        for (double expectedValue : expectedValues)
+        {
+            assertThat(scores.getValue(current), closeTo(expectedValue, E));
+            current = current.minus(1);
+        }
+    }
+
+    @Test
     public void test_getValues()
     {
         toggleRepetitions(0, 20);
@@ -196,5 +239,12 @@ public class ScoreListTest extends BaseUnitTest
 
         for (int i = from; i < to; i++)
             reps.toggle(today.minus(i));
+    }
+
+    private void addSkip(final int day)
+    {
+        RepetitionList reps = habit.getRepetitions();
+        Timestamp today = DateUtils.getToday();
+        reps.toggle(today.minus(day), Checkmark.SKIP);
     }
 }


iSoron:
Thank you for the pull request, @KristianTashkov. I will put some more thought into this and give you some feedback.

chennemann:
As stated in the other discussions before, I'm with @KristianTashkov on this. I do agree that skipping days should neither punish nor reward the user.

iSoron:
Hi @KristianTashkov and @chennemann, here is how I see this issue:

1. I agree that skip days are not equivalent to explicit checkmarks; that causes strange issues. The one that bothers me the most is that adding a skip day causes implicit checkmarks to appear. Thinking more about this issue, I now think skip days are equivalent to implicit checkmarks. The way I see it, before #630, only the app itself could add skip days, based on the habit frequency. After #630, the user can also add skip days manually. I will push a commit soon making `SKIPPED` and `CHECKED_IMPLICITLY` essentially the same the same thing, and having similar visual representation, differing only by color (automatic is grey, manual is colorful). I'd be interested in hearing your opinion on this. My next comments will assume that you agree with this interpretation. That is, I will treat `SKIPPED` and `CHECKED_IMPLICITLY` as the same thing.

2. **Beginning and ending a streak with skipped days.** Since the first public version of the app, skips have always contributed to the length of a streak, even if they appear at the end of the interval. Since #124, streaks can also begin with skips. *I am not planning to change this long-standing app behavior.*

3. **Skipped days being counted as successful days in streak.** The number indicates the length of a streak in days. As I wrote in #37: the idea for the current way of counting is that it is consistent across all habits. If you see the number "14", you know that you have been doing the habit perfectly for two weeks; if you see "365", you have been doing it for a year, and so on, no matter what is the frequency of the habit. By counting repetitions, the length of time becomes unclear for habits with complex schedule. *I don't believe there's enough user demand to change this feature.*

4. **Skipping days changes the score.** This has also been discussed at length. Initially, skips caused the score to slowly decrease (#463). More recently, skips caused the score to slowly increase. I would accept a pull request making the scores fixed during skip days, but only if this does not affect the computation for non-daily habits (that is, it should still take the same number of days to achieve the same score). The new score formula should also not distinguish between automatic (`IMPLICIT_CHECKMARK`) or manual (`SKIPPED`) skips.

KristianTashkov:
While I see where you are coming from, I don't agree that implicit checks and skip days are the same thing. Going back to our discussion about possible states of a habit:
1) Habit is not due today -> implicit check (this only makes sense for non-daily habits)
2) Habit is due today -> which can be either unchecked or checked
3) Habit is not valid for today -> what I believe skip days are.

I agree there are a lot of similarities between 1) and 3) in the way they affect scores/streaks. One thing I'm very confidently against is the newest change you pushed where implicit checks are shown with a `-`. I have seen so many threads where people like the full checkmarked view of a habit and this completely kills that and makes it look like something wrong happened.

On the streaks front: I generally agree with the opinion you have given in different threads that streaks are a bad indicator of progress, so my opinion on this isn't very strong. The way I interpret the current streak number is "days where I have not failed the habit in a row". This to me includes both case 1) and 2), whenever I had the option to fail, I didn't. I'm okay with skipped days being added to that count, but it does feel wrong that a streak of fully skipped days can be shown as better than one where you did it every day, which is why I introduced the change.

On the scores front: I'm 100% sure in my mind that skipped days should be neutral in terms of score. For implicit checks I'm really not sure. I agree they shouldn't decrease your score. Making them neutral seems like an okay solution. I kind of like that even on "break" days you still get a little curve up, a little motivation that you are on a streak and still improving.

I guess my biggest argument for them being different is that it feels wrong to be considered successful on days where the habit isn't valid. I see this most strongly for work habits. I like that on weekends I have a `-` because it's not a valid day and it always annoyed me that if I do my workday habit 5 times, the weekend gets an implicit check, if I don't I see it gray as if failed. Also it's nice to visually see weeks where you were on a vacation and see the score frozen for them.

chennemann:
I don't have much to add to what @KristianTashkov said. 
@iSoron I think regarding your points 2-4 we are on the same page and can get to some kind of consensus. 

As @KristianTashkov explained I don't think skipped days and skipped days are the same thing. But I just noticed (while writing this) a subtle point in your argument that we (please correct me if I'm wrong here @iSoron & @KristianTashkov) missed previously.


> The one that bothers me the most is that adding a skip day causes implicit checkmarks to appear.

So if I have a 5 times a week habit and I do it monday to thursday and skip it on friday then saturday and sunday will be implicitely checked. Is that what you were referring to @iSoron?

iSoron:
> So if I have a 5 times a week habit and I do it monday to thursday and skip it on friday then saturday and sunday will be implicitely checked. Is that what you were referring to @iSoron?

Before c846dfc75a89c142f833e6920ef84e5e81dfee8d, a skip was equivalent to an explicit checkmark, so adding 5 skips (Mon-Fri) would automatically generate 2 extra skips (Sat-Sun). After c846dfc75a89c142f833e6920ef84e5e81dfee8d, skips are equivalent to implicit checkmarks, so adding 5 skips would not generate any new skips automatically. I consider the new behavior correct.

In your example, if you add checkmarks from Monday to Thursday then add a skip on Friday, the app will not add automatic skips on Saturday or Sunday. You can still, however, add them yourself manually. Manual and automatic skips have the same effect on scores and streaks, so it's just a matter of convenience.

> 1. Habit is not due today -> implicit check (this only makes sense for non-daily habits)
> 2. Habit is due today -> which can be either unchecked or checked
> 3. Habit is not valid for today -> what I believe skip days are.

I see it as follows:

1. `YES`: The habit has been completed.
2. `SKIP`: The habit is not applicable (e.g. it's not due; the user is sick; it's a holiday; etc).
3. `NO`: The habit was applicable, but was not completed; a lapse.

Before, we could only skip days if the habit was not due, according to the app's internal scheduling algorithm. The skip had to be added automatically by the app, and there was no way to do it manually. Now we can add skips manually for whatever reason.

KristianTashkov:
I see what you mean, but I don't agree with it. Especially visually `not due` and `not applicable` should be different in my opinion, and changing the homescreen to have a lot of `-` instead of checks feels like a step back for motivation.
It's your app and if this is the direction you are going with you can close this PR.

On a separate note, I'm sad to say contributing here has been a very negative experience. There are so many features that users have demanded for literally years and I understand that you may not have the time to implement them yourself, but you are not using contributors effectively. Even though you answer in issues that you are okay with PRs and would like a feature built, when someone does build it you effectively rewrite it the way you envisioned it (which was never mentioned before) ignoring their contribution. I suggest you write a clear guidance for contributors on how you want the most wanted features implemented, so someone can write the code and you just review. Offloading any desired changes to them is also preferred than you rewriting it yourself, it will save you time and it will teach contributors how to write better code for their PRs so they are more quickly merged in the future. I was really looking forward to contributing a lot of features here (have a whole kanban board with them for myself), but managing 3+ PRs for months and rebasing on top of your changes while the PRs stay unreviewed has been unpleasant and unproductive. At this point I will just continue implementing features for myself and if they are easily ported to the dev branch I will try to open a PR in the hopes that you will find it useful and merge it for everyone some day.

iSoron:
Hey @KristianTashkov, I'm sorry to hear it was such a negative experience to you, but thank you for the feedback. I will keep it in mind when reviewing future PRs from you and from other collaborators. I do hope you stick around, as you have provided some good contributions to the project.

adrienbeau:
> I see what you mean, but I don't agree with it. Especially visually not due and not applicable should be different in my opinion, and changing the homescreen to have a lot of - instead of checks feels like a step back for motivation.

I would usually just react with an emoji, but since your comment contains two very separate parts, I have to highlight this one and say that I very much agree with you on this point. While implicit checks and skips are very similar, and may behave the same in many parts of the app (streaks for example), they are different in other aspects you have explained better than I could.

On the specific topic of visual rendition, I also like the current "grey check" of implicit checks, and think skips should have a different symbol such as a dash.

iSoron:
I have reverted the visual changes to implicit checkmarks in 5fa3f41. They are now shown as grey checkmarks, like before. I still think implicit checkmarks are much closer in concept to skips, but, since this visual change seems to be controversial, let's keep the current app behavior.

Other than this, is there still any other point of contention?

iSoron:
Hey everyone, I would also appreciate your opinion on the following scenario. Suppose the user has a 5-times-per-week habit. They complete the habit from Monday to Thursday, then skip Friday. What should happen to the weekend?

Currently, the app will not add implicit checkmarks to the weekend, since there has not been 5 explicit checkmarks in the past 7 days. In addition:
1. The user can add skips manually to the weekend, but now these days will look different from the previous weekends, so, in a sense, the user may still feel punished for skipping Friday, although this doesn't really affect the scores or streaks. *This was my main motivation for making implicit checkmarks and skips visually identical*.
2. Recall that skips currently increment the score. If we make skips simply copy the previous score, then the punishment is no longer just visual; skipping any day of the week causes three days of score stagnation, instead of just one. *This is my main motivation for making implicit checkmarks and skips identical for the purposes of score calculation*.
3. If we make skips count when generating implicit checkmarks, then skipping Mon-Fri will generate two checkmarks during the weekend. If, in addition, we make skips copy the previous score, then the score will remain the same for five days, then increase for two days during the weekend, which I would say is very unexpected.

KristianTashkov:
> Hey everyone, I would also appreciate your opinion on the following scenario. Suppose the user has a 5-times-per-week habit. They complete the habit from Monday to Thursday, then skip Friday. What should happen to the weekend?

I think the problems you outline happen because we are not making a distinction between regular and irregular habits. To show the difference I will give examples:
1) If you have a work habit that is only possible Mon-Fri, this is a strict habit and should be an every "Mon-Fri" habit, not a 5 out of 7 days habit.
2) An irregular habit might be, work out 5 times every week, but not have a strict schedule to it. This way the app can automatically fill the gaps with successes for the 2 days you didn't do it.

I'm trying to address this problem in https://github.com/iSoron/uhabits/pull/633.

For the specific problem you are outlining this will look like this:
1) If the habit you outlined is considered strict and set as inactive during the weekend, all weekends will be skipped (both those with 5 or 4 occurrences), so everything is consistent. If the habit is not strict, then it's correct to show the weekend as unchecked and remind you to do the habit, since you still can get a 5/7 week.
2) I still think pausing the score for skips is the right way to go. If the habit is strict, every weekend pauses the score, which makes sense because it's days where the habit doesn't make sense to succeed or fail. If the habit is not strict, getting punished or rewarded for days it wasn't possible to do (e.g. sick days) doesn't make sense.
3) I don't think skips should generate implicit checkmarks for the reasons you mentioned, they are not successful repetitions for either regular or irregular habits.

I think the skip discussion is a bit unrelated to `YES_AUTO` to be honest and I am a lot more clear on how `SKIP` should work, than I am about `YES_AUTO`.
`YES_AUTO` only makes sense for irregular habits to automatically mark the non-due time successful for you, strict habits with a weekly schedule should be yes/no or skip for all the days. I like how the app handles it today with automatically converting the irregular habit to a regular looking one by filling in the gaps for you. There is an argument to be made that these "non-due" days are essentially skipped (which is what @iSoron is defending), but having the whole week look like a success, and keep the skipped visual for specific events feels more satisfying. In general I have other problems with how we handle irregular habits in the app that I have expressed in a different issue, but it might be out of scope for the current discussion.

KristianTashkov:
> Other than this, is there still any other point of contention?

I still have two:
1) Skip days should pause score.
2) Skip days should be excluded from streaks (but don't break them). I feel like I can't win this fight but I'm giving it a last try before giving up and just resort to keep pulling it in my personal version when updating from dev.

image
image
This is a work habit that I have been keeping track of for a while and the screenshot is from my personal version with the changes I'm rooting for. Every weekend I mark it as a skip (because the habit can't be completed then). The last two weeks I was on vacation so I marked all of them with skips as well.
1) The score for the past 2 weeks is paused, I was on vacation and haven't improved or regressed on this habit.
2) My current streak is shown as 5 days, because I have 5 work days in a row since the last time I failed the habit. The current dev will instead show a streak of 24 which will dwarf the best streak I have with a meaningless data point, I wasn't even working during that time. Yes the streak number is no longer showing actual days between the start end, but I don't need it to, it already shows the actual dates. Instead it shows a meaningful data point to me, how many work days did I do the habit in a row. Weekends/sick days/vacations shouldn't matter in that number.

chennemann:
@iSoron thanks for the clarification of the problem. I do agree that with the current state of the application this would probably be the best solution. @KristianTashkov has a point though with the concept of regular and irregular habits. For me skipping a day of a habit (for example refuse sweets) is only applicable at all when I won't be able to complete it this week anyways. So if it is a 5/7 habit and I'm at 4/7 I would not necessarily skip Friday or Saturday because I know I can still complete it on Sunday. I would only skip a day (or in that case Friday to Sunday) if I have a valid reason why I could not do the habit. But since some habits are bound to a specific day the addition of applicable days as suggested in #633 would (as @KristianTashkov mentioned) mitigate the mentioned problems.

iSoron:
@KristianTashkov wrote:
> This is a work habit that I have been keeping track of for a while and the screenshot is from my personal version with the changes I'm rooting for. Every weekend I mark it as a skip (because the habit can't be completed then).

Is there any reason why you are adding these skips manually, instead of letting the app add them for you? If you have a 5-times-per-week habit and you actually complete it 5 times, the app will add automatic checkmarks during the weekend.

> I think the problems you outline happen because we are not making a distinction between regular and irregular habits.

I don't think this is the root cause. Suppose I have a habit that needs to be done 2 times per week, but I can only do it on a few particular days (say, Monday, Wednesday or Friday). This week, I did it on Monday, but I can't do it Wednesday or Friday for a good reason, so I wouldn't like the app to punish me. If I enter `SKIP` from Wednesday to Sunday and the app freezes my score for 5 days, that's quite a punishment (5 days of score stagnation for a single skipped day). See below for an illustration. If we make **both** `YES_AUTO` and `SKIP` freeze the score, then I guess it could work, but this would also bring its own set of issues.

screen

chennemann:
@iSoron if we had the concept of specific days this would not be the case because for 3 out of the 5 skips the score would have been frozen anyways because those days are not eligible for the habit. So it might not be the cause but imo a rather elegant solution.

KristianTashkov:
> Is there any reason why you are adding these skips manually, instead of letting the app add them for you? If you have a 5-times-per-week habit and you actually complete it 5 times, the app will add automatic checkmarks during the weekend.

Exactly because of the problem we are discussing. These "automatic" checkmarks only work if you do it 5 times, if I fail even a single day, the whole weekend looks different than the rest. For me every weekend is skips, the habit isn't active then.

> I don't think this is the root cause. Suppose I have a habit that needs to be done 2 times per week, but I can only do it on a few particular days (say, Monday, Wednesday or Friday). This week, I did it on Monday, but I can't do it Wednesday or Friday for a good reason, so I wouldn't like the app to punish me. If I enter SKIP from Wednesday to Sunday and the app freezes my score for 5 days, that's quite a punishment (5 days of score stagnation for a single skipped day). See below for an illustration. If we make both YES_AUTO and SKIP freeze the score, then I guess it could work, but this would also bring its own set of issues.

First, why do you say the app is punishing you when the score is frozen? If you have a "good" reason why the habit couldn't be done for Wednesday or Friday, it's not a fail so we shouldn't decrease the score, but you also didn't do the habit, so no score bonus is warranted.
Secondly, with the concept of active days this habit will look like this:
image

Every day that is not mon/wed/fri is always a skip. Then you only add two other skips for the special occasion of the habit not being completable for the rest of the week.

iSoron:
@chennemann @KristianTashkov In the last screenshot shared by Kristian, could you please clarify in which of these days is the score increasing, decreasing or frozen?

chennemann:
All mondays, wednesdays and Fridays except the 2.9 and 4.9. 

At least that's my understanding

KristianTashkov:
For a habit only possible on Mon/Wed/Fri, every other day the score is frozen (Tue/Thu/Sat/Sun).
02/09 and 04/09 are special occasions where the score is manually frozen (skipped explicitly from the user).
The rest is the same as it is normally in the app (any type of checks increase the score, missed repetitions decrease it)

iSoron:
1. Forgetting manual skips for a moment and focusing only on score calculation, what if the user has exactly the same habit as before (2 times per week) but there is now only one invalid day of the week (say, Sunday). Below is a screenshot where dashed-light boxes are `SKIP_AUTO` (this is a new concept) and solid-light boxes are our current `YES_AUTO`. Is the score only frozen on Sundays? It keeps increasing in all remaining `YES_MANUAL` and `YES_AUTO` days?

2. Suppose that the user changes the invalid days of the week after performing the habit for a while. Does this affect previously calculated checkmarks and scores? How do you handle that?

screen

KristianTashkov:
> Is the score only frozen on Sundays? It keeps increasing in all remaining YES_MANUAL and YES_AUTO days?
In the current proposal, yes. The more we talk about it the less convinced I become that score should increase on `YES_AUTO` days but even today it doesn't really move that much anyway so it's negligible.

> Suppose that the user changes the invalid days of the week after performing the habit for a while. Does this affect previously calculated checkmarks and scores? How do you handle that?

This is a problem we already have in the app, when you change the habit to a 3 per week (you are getting better at it and want to bump it from just 2 per week), and you get score decreases for all past days. Both problems have solutions (tracking history of frequency/active days changes for example) but that is a separate discussion all together.

iSoron:
@KristianTashkov I could see your proposal to freeze scores working if we did the following:

1. Rename `SKIP` to `SKIP_MANUAL` and `YES_AUTO` to `SKIP_AUTO`.
2. Make both `SKIP_MANUAL` and `SKIP_AUTO` freeze the score.
3. Make `YES_MANUAL` increase score of all habits by the same amount, regardless of habit frequency.

The one drawback I see with this solution is that, if the user happens to perform the habit more frequently than the target frequency, then the score will increase that much faster. This allows the user to "catch up" and reach a high score much faster than it would be otherwise possible. The app used to allow this in its earlier versions and I never liked it, but I guess I can live with it.

If we move in this direction, then I will probably insist in making all skips visually the same, since they *would be* the same for all purposes. This would also allows us to have a more consistent UI where the symbol indicates what's happening to the score and the color represents the source (color=manual, grey=automatic).

Assuming we move in this direction, what benefits do we get from the concept of "invalid days"? If the user does perform the habit 5 times in a week, the app automatically adds `SKIP_AUTO`. If the user performs the habit 4 times a week, they can still manually add `SKIP_MANUAL` if it bothers them. Both would have the same effect. It particularly never bothered me, and it's also not something I have heard from users.

@KristianTashkov wrote:
> This is a problem we already have in the app

I agree, but this proposal would make the problem worse, so I would like to see it solved first (#270). I am also not sure that issue could be solved with a simple UI. We would need not only to keep track of the changes in our database, but also allow the user to see and edit those change points in an intuitive way.

chennemann:
> Assuming we move in this direction, what benefits do we get from the concept of "invalid days"?

With your proposed solution we won't get an additional benefit. But the idea is/was that there is a distinction between SKIP_AUTO and YES_AUTO in that sense that the algorithm would add SKIP_AUTO for invalid days in a fixed habit and YES_AUTO for implicitely completed day (the way it is currently). 

We also need two different states (SKIP_AUTO and YES_AUTO) because we need to distinguish between invalid days and valid days that are implicitly completed. 

Yet again I want to ask for your feedback on my proposal regarding the extraction of the source of a habit (manual/auto) to minimize the checkmark states we need to maintain. (I guess my comments got lost somewhere :)) 
I'll just copy my proposal from here:
https://github.com/iSoron/uhabits/pull/610#issuecomment-679233117

### Checkmark States

> 4. I'm concerned about the number of checkmark states we now have. Before this pull request, there were only three (CHECKED_EXPLICITLY, CHECKED_IMPLICITLY and UNCHECKED). Given the two comments above, I think this feature can be implemented with just four states (CHECKED_EXPLICITLY, CHECKED_IMPLICITLY, UNCHECKED and SKIPPED). I suggest merging all FAILED_* into UNCHECKED. Repetitions for boolean habits could now have either CHECKED_EXPLICITLY or SKIPPED states.

I do agree that we should be careful with adding new Checkmark states since they make the whole codebase more complicated and the process harder to reason about. However I do think it is very valuable to have a state for each actual state :D

- Unchecked (when nothing is entered)
- Checked (when the habit is completed for the day)
- Skipped (when the habit is skipped for the day)
- Failed (when the habit is failed for the day)

I think a big reason why adding more states is a problem in the first place is the encoding of the source (implicitly/explicitly) into the states which then multiplies the states exponentially.
To solve this I would propose to have an additional boolean flag that indicates whether it is an explicit input or the result of our computation.

KristianTashkov:
I'm with @chennemann on this.
> Assuming we move in this direction, what benefits do we get from the concept of "invalid days"?

1) It's automatic, you don't have to manually click `SKIP_MANUAL` for each work habit you have for both days of the weekend. You don't get a reminder for those days (overlaps with the current reminder days). Not to mention with your solution if you did a work habit 5 times, the weekend will be colorless `-`, if you did it 4 times you will have to manually add colored `-`, making it look different for no reason.
2) Your solution goes back to the visual representation where your home screen is a few checks with a lot of `-`. I thought we already moved away from that.
3) Even if `YES_AUTO` and `SKIP_MANUAL` have the same behavior score/streak-wise, I still think they should have a different visual representation because they mean different things.

> Make YES_MANUAL increase score of all habits by the same amount, regardless of habit frequency.

Wouldn't that make 1 per week habits 7 times slower to progress than every day habits, which sounds like a bad idea and a pretty big change to the current behavior.

The way I see it is:
1) `YES` / `NO` / `SKIP` are the different states. `YES` always increases score, `NO` always decreases, `SKIP` is freezing
2) Each of them has a `AUTO` or `MANUAL` equivalent (which can be a separate variable for simplicity like @chennemann is proposing)
3) Since `SKIP_AUTO` affects frequency, it needs to be accounted in score calculations so habits of all frequencies and all invalid days states progress by the same amount. For example if you have an every day habit with 2 invalid days, it should be equivalent to 5/7 habit.

Technically we can skip point 3 (and I don't like how convoluted it is), if we remove `SKIP_AUTO` and put `YES_AUTO` in the same place for invalid days, but then you don't get the nice skip visual for invalid days.


> We would need not only to keep track of the changes in our database, but also allow the user to see and edit those change points in an intuitive way.

Having the list is one thing which isn't hard to achieve either UI or code wise. This gets us 99% there. Having to edit the list sounds like a separate convenience feature.

iSoron:
@KristianTashkov wrote:
> Wouldn't that make 1 per week habits 7 times slower to progress than every day habits, which sounds like a bad idea and a pretty big change to the current behavior.

That's the current app behavior.

@chennemann wrote:
> Yet again I want to ask for your feedback on my proposal regarding the extraction of the source of a habit (manual/auto) to minimize the checkmark states we need to maintain. 

I think we are already moving in that direction with `YES_MANUAL` and `YES_AUTO`. As for splitting this into two fields, sure, we could do it if it simplifies the code base.

iSoron:
This issue started with a proposed solution to #88, then it morphed into something else. At the moment, it seems both @chennemann and @KristianTashkov are proposing a new concept of "invalid days", where the app automatically adds either checkmarks or skips to these days. I'm open to it, but I would like to understand it better. I am particularly interested in understanding how would this proposal fit with our existing concept of non-daily habits and the score formula.

For each one of the habits below, and for each day from July 30 until Sep 7, could you guys please let me know: 
1. What is the checkmark state? `YES_MANUAL`, `YES_AUTO`, `SKIP_AUTO`, `NO` or something else.
2. What is the score for the habit for each day? Please use actual numbers here, not just "it increases" or "it stays the same". I am trying to understand your proposed formula too.

The habits are similar to the ones I mentioned before:

* **Habit 1:** I should do it twice a week, and I can do it every day. I did it on Aug 3, 5, 10, 14, 17, 19, 24, 28. There are no other successful repetitions.

* **Habit 2:** I should do it twice a week, but it's impossible for me to do it on Sunday. I did it on the same days as before.

Note that, from Aug 31 to Sep 7, we start having `NO` states. Please let me know what are the scores for these days too. I want to understand how your proposed formula works in that situation. The table below shows the current app behavior for both habits (there is no distinction since the concept of "invalid days" does not exist):

Date | Score | State
-- | -- | --
2020-09-07 | 0.3074 | NO
2020-09-06 | 0.3121 | NO
2020-09-05 | 0.3169 | NO
2020-09-04 | 0.3218 | NO
2020-09-03 | 0.3267 | NO
2020-09-02 | 0.3317 | NO
2020-09-01 | 0.3368 | NO
2020-08-31 | 0.342 | NO
2020-08-30 | 0.3472 | YES_AUTO
2020-08-29 | 0.3372 | YES_AUTO
2020-08-28 | 0.327 | YES_MANUAL
2020-08-27 | 0.3167 | YES_AUTO
2020-08-26 | 0.3062 | YES_AUTO
2020-08-25 | 0.2956 | YES_AUTO
2020-08-24 | 0.2848 | YES_MANUAL
2020-08-23 | 0.2738 | YES_AUTO
2020-08-22 | 0.2626 | YES_AUTO
2020-08-21 | 0.2513 | YES_AUTO
2020-08-20 | 0.2398 | YES_AUTO
2020-08-19 | 0.2282 | YES_MANUAL
2020-08-18 | 0.2163 | YES_AUTO
2020-08-17 | 0.2043 | YES_MANUAL
2020-08-16 | 0.1921 | YES_AUTO
2020-08-15 | 0.1797 | YES_AUTO
2020-08-14 | 0.1671 | YES_MANUAL
2020-08-13 | 0.1543 | YES_AUTO
2020-08-12 | 0.1413 | YES_AUTO
2020-08-11 | 0.1281 | YES_AUTO
2020-08-10 | 0.1147 | YES_MANUAL
2020-08-09 | 0.1011 | YES_AUTO
2020-08-08 | 0.0874 | YES_AUTO
2020-08-07 | 0.0733 | YES_AUTO
2020-08-06 | 0.0591 | YES_AUTO
2020-08-05 | 0.0447 | YES_MANUAL
2020-08-04 | 0.03 | YES_AUTO
2020-08-03 | 0.0151 | YES_MANUAL
2020-08-02 | 0.0591 | YES_AUTO
2020-08-01 | 0.0447 | YES_AUTO
2020-07-31 | 0.03 | YES_AUTO
2020-07-30 | 0.0151 | YES_AUTO

KristianTashkov:
I will give it a bit more thought and give you a table after I implement it to see how it looks.

In the meantime I agree that the discussion got morphed a bit and we should probably split it. The concept of invalid days was proposed as a solution to the problem where weekends are either `NO` if you do 4 repetitions or `YES_AUTO` if you do 5, for a 5/7 habit. This is a bit different than explicit skip days so maybe we should split off that discussion in https://github.com/iSoron/uhabits/pull/633. If you take the lowest form of that feature, it just adds an automatic way to add `YES_AUTO` beside our frequency calculations, no need to change score calculations for it.

In terms of this current thread about explicit skips, getting it down to the lowest idea is that `SKIP` in its current form, should freeze the score, because it's a special event and different than the regular `YES_AUTO` (even visually).
Trying to merge these two ideas into something more complex and cohesive requires thinking about how we generally handle non-daily habits. I think to better discuss this, we need to gather a list of actual habits people track and their settings, so we can discuss it with actual examples (like user stories), and see what makes sense in terms of score for each.

P.S. Does it make sense to start something like a discord server/slack so we can discuss ideas more easily. Async comment systems become a bit tiresome for complex ideas that we can't quickly get on the same page.

iSoron:
For reference, here is my proposed solution from an earlier post, renaming `YES_AUTO` to `SKIP_AUTO` and freezing the score. When the status is `YES`, the new score is `0.94807 * prev_score + 0.05192`. When the status is `SKIP`, the score stays the same. When the status is `NO`, we need to tweak the formula, so that the score doesn't decay so rapidly; here, I propose using the formula `0.9849 * prev_score + 0.0151`. Note that this table matches the previous table exactly on `YES` and `NO` days. 

The magic numbers are `0.94807 = (1/2)^(2/7/13)`, `0.05192 = 1 - 0.94807`, `0.9849 = (1/2)^(1/13)` and `0.0151 = 1 - 0.9849`. The fraction `2/7` comes from the habit frequency.

Date | Score | State
-- | -- | --
9/7/20 | 0.3074 | NO
9/6/20 | 0.3121 | NO
9/5/20 | 0.3169 | NO
9/4/20 | 0.3218 | NO
9/3/20 | 0.3267 | NO
9/2/20 | 0.3317 | NO
9/1/20 | 0.3368 | NO
8/31/20 | 0.3420 | NO
8/30/20 | 0.3472 | SKIP_AUTO
8/29/20 | 0.3472 | SKIP_AUTO
8/28/20 | 0.3472 | YES
8/27/20 | 0.3115 | SKIP_AUTO
8/26/20 | 0.3115 | SKIP_AUTO
8/25/20 | 0.3115 | SKIP_AUTO
8/24/20 | 0.3115 | YES
8/23/20 | 0.2738 | SKIP_AUTO
8/22/20 | 0.2738 | SKIP_AUTO
8/21/20 | 0.2738 | SKIP_AUTO
8/20/20 | 0.2738 | SKIP_AUTO
8/19/20 | 0.2738 | YES
8/18/20 | 0.2340 | SKIP_AUTO
8/17/20 | 0.2340 | YES
8/16/20 | 0.1921 | SKIP_AUTO
8/15/20 | 0.1921 | SKIP_AUTO
8/14/20 | 0.1921 | YES
8/13/20 | 0.1478 | SKIP_AUTO
8/12/20 | 0.1478 | SKIP_AUTO
8/11/20 | 0.1478 | SKIP_AUTO
8/10/20 | 0.1478 | YES
8/9/20 | 0.1011 | SKIP_AUTO
8/8/20 | 0.1011 | SKIP_AUTO
8/7/20 | 0.1011 | SKIP_AUTO
8/6/20 | 0.1011 | SKIP_AUTO
8/5/20 | 0.1011 | YES
8/4/20 | 0.0519 | SKIP_AUTO
8/3/20 | 0.0519 | YES

KristianTashkov:
Here is my proposition for least amount of changes that gets the desired effect. Two changes from today's dev:
1) frequency variable calculation to become:  `frequency_nom /  (frequency_den * active_days / 7)`.
2) `SKIP_AUTO` copies the score from the previous day.

The idea behind is to scale the days according to how many will be skipped so you get the same result in the end.
* 5/7 habit with all valid days, should progress the same way as a 7/7 habit with 2 invalid days.
* 2/7 habit with 1 invalid day, should be the same as a 2/6 habit.

If you compare a 1/7 habit before, with a "every sunday" habit, the score will be the same, but while in master you get 1/7 of the score each day, with the invalid days you will get the whole bump in one day. Equivalently since there is only a single day where you can fail, if you do fail on sunday, it will be equivalent to you failing the entire week. (If the habit is really possible ONLY on sunday, then these scenarios are the same in terms of score since you would have failed 7 days in a row, just more sudden)


The table looks like this, with the score on every sunday matching what it is in dev, but the interval between is progressing faster to account for the sunday that pauses the score.


Date | Score | State
-- | -- | --
9/7/20 | 0.3066 | NO
9/6/20 | 0.3121 | SKIP_AUTO
9/5/20 | 0.3121 | NO
9/4/20 | 0.3177 | NO
9/3/20 | 0.3234 | NO
9/2/20 | 0.3292 | NO
9/1/20 | 0.3351 | NO
8/31/20 | 0.3411 | NO
8/30/20 | 0.3472 | SKIP_AUTO
8/29/20 | 0.3472 | YES_AUTO
8/28/20 | 0.3355 | YES_MANUAL
8/27/20 | 0.3236 | YES_AUTO
8/26/20 | 0.3115 | YES_AUTO
8/25/20 | 0.2991 | YES_AUTO
8/24/20 | 0.2866 | YES_MANUAL
8/23/20 | 0.2738 | SKIP_AUTO
8/22/20 | 0.2738 | YES_AUTO
8/21/20 | 0.2608 | YES_AUTO
8/20/20 | 0.2475 | YES_AUTO
8/19/20 | 0.2340 | YES_MANUAL
8/18/20 | 0.2203 | YES_AUTO
8/17/20 | 0.2063 | YES_MANUAL
8/16/20 | 0.1921 | SKIP_AUTO
8/15/20 | 0.1921 | YES_AUTO
8/14/20 | 0.1776 | YES_MANUAL
8/13/20 | 0.1628 | YES_AUTO
8/12/20 | 0.1478 | YES_AUTO
8/11/20 | 0.1325 | YES_AUTO
8/10/20 | 0.1170 | YES_MANUAL
8/9/20 | 0.1011 | SKIP_AUTO
8/8/20 | 0.1011 | YES_AUTO
8/7/20 | 0.0850 | YES_AUTO
8/6/20 | 0.0686 | YES_AUTO
8/5/20 | 0.0519 | YES_MANUAL
8/4/20 | 0.0349 | YES_AUTO
8/3/20 | 0.0176 | YES_MANUAL

Alternatively you can incorporate your proposition into this as well. We can still make `YES_AUTO` freeze score and make the score of `YES_MANUAL` static. That way even in irregular habits you get an increase of score only when you do it, and no increase the other days. But I personally don't like it as much:
1) Changes existing behavior of off-days increasing the score that people are already used to.
2) Makes `YES_AUTO` and `SKIP` look the same way, which I don't agree should be the case, because they are different things.

iSoron:
Thanks, your proposal is becoming more clear to me. I still have a few questions:

1. How does this interact with the generation of `YES_AUTO` checkmarks? For example, suppose I have a habit that I should perform every three days, but I cannot do it on Sundays. If I perform it on Saturday, what days get `YES_AUTO` and what days get `SKIP_AUTO`?

2. Some combinations of habit frequency and invalid days seem to be invalid. For example, it doesn't make sense to have a "daily habit" that I cannot perform on weekends; or to have a "5 times per week habit" that I cannot perform on Saturday, Sunday and Monday (3 invalid days); or to have a "20 times per month habit" that I can only perform on Mondays. What is your proposal here? Allowing the user to enter these invalid combinations anyway (this seems to be the case in your branch) or disallowing them?

KristianTashkov:
1) I don't think it interacts in any way, but I also really dislike the current implementation of ``YES_AUTO` anyway. To my understanding `YES_AUTO`  has two major purposes:
 * Fill the holes between the first and last repetition in the desired interval to be marked as successful. (`SKIP_AUTO` doesn't need any interaction here, only that it overrides the `YES_AUTO`)
 * If you do your last repetition before the end of the interval to give you a break of a couple of days before you should start the habit again. (One of those days being a skip day doesn't really change anything in terms of how long the break is, so again it shouldn't affect it. Only interaction is that `SKIP_AUTO` overrides the `YES_AUTO` for that day.).
In terms of score the formula should have scaled the value of all `YES_*` to account for the fact that some days are skips and don't increase the score.

2) For now I was mostly trying to show how the score works and the current code handles this in the backend by treating all of those cases as if it they were set as "every day" habits, since that's what they equate to. We can try to improve the UI to show that this is what was submitted. For example we can turn the text into `Every active day` for those cases.

iSoron:
> but I also really dislike the current implementation

Please let's keep this discussion focused on your "invalid days" proposal. If your proposal includes the removal of `YES_AUTO`, then make this clear and please clarify how would "invalid days" replace it. If it includes changing `YES_AUTO` behavior in any way, please clarify what exactly would change.

> For now I was mostly trying to show how the score works and the current code handles this in the backend by **treating all of those cases as if it they were set as "every day" habits, since that's what they equate to**.

Sorry, but you lost me again. So if the habit has at least one invalid day, are you proposing that the frequency becomes fixed to "daily"? This is inconsistent with your previous posts where you showed us an alternative score formula that *did* include habit frequency.

> For example we can turn the text into Every active day for those cases

I would like to see either mocks or a working implementation of this GUI before I can give any comments. This is something that can easily become very confusing.

My opinion on this proposal so far, unfortunately, is that it has not been well thought out; the proposal seems to change completely after each question I ask and it does not seem to play well with our existing features (specifically, non-daily habits).

KristianTashkov:
> Sorry, but you lost me again. So if the habit has at least one invalid day, are you proposing that the frequency becomes fixed to "daily"? This is inconsistent with your previous posts where you showed us an alternative score formula that did include habit frequency.

That is not what I'm saying at all. I am saying that if you have X active days in a week/month/etc, and you specify a frequency that is more than X, it essentially means that the habit has to be done on every day that it is valid. For example if you have a habit that isn't possible on Saturday or Sunday, specifying 5 per week, 6 per week or every day is essentially the same thing, you don't get `YES_NO` days.

> I would like to see either mocks or a working implementation of this GUI before I can give any comments. This is something that can easily become very confusing.

This was just a suggestion on the top of my head, there are probably better UIs for this. It makes sense to warn the user that if he has 3 invalid days, he can't possibly do the habit 6 times every 7 days. To be honest I don't expect a lot of users will fall into that hole since it's fairly obvious and as long as the app doesn't behave in a weird way, it should be okay. Better UI is of course preferable.

> My opinion on this proposal so far, unfortunately, is that it has not been well thought out; the proposal seems to change completely after each question I ask and it does not seem to play well with our existing features (specifically, non-daily habits).

I mentioned this before but we combined a couple of discussion in the same thread. I am proposing two separate things that haven't changed much since I initially proposed them.

1) This PR started as a desired change to `SKIP`/`SKIP_MANUAL` repetitions. It was to freeze the score when the user manually picks a SKIP and to exclude it from streaks (the streaks part seems to have less support from the community). You raised some questions of how we handle a specific problem for work habits that the user has to manually skip weekends and how it makes a difference to the current `YES_AUTO` behavior. My answer to that was that I feel like it's a separate question and a problem with our `YES_AUTO` behavior for habits that are never possible on specific days (the current workaround of specifying 5/7 and rely on `YES_AUTO` for the invalid days). This prompted my second suggestion:
2) https://github.com/iSoron/uhabits/pull/633 which was to be able to specify explicit invalid days and not rely on `YES_AUTO`. The latest discussions should have been moved there when I initially linked it but we kept writing here.

KristianTashkov:
Updated the PR to make it clear the change this PR is trying to introduce.

iSoron:
If the proposal is just to keep the score fixed on `SKIP`, I am fine with it. I also like your implementation, so I will go ahead and merge it. Thank you for the PR and for the discussion, @KristianTashkov.

As for #633, I will open a new issue about it and we can move the discussion there.

iSoron:
Please see #641

PR #631, opened 2020-09-01
Fix habit selection menu item background color

KristianTashkov:
When using a dark theme the menu for actions of selected habits has a white background on white text.
image
After this PR:
image


iSoron:
Thanks, merged!

PR #633, opened 2020-09-04
Make reminder days into active days for the habit

KristianTashkov:
Hey @iSoron, wondering what you feel about a change like this.
I don't have the exact issue at hand but I remember people wanting to specify which days should the habit be done. The answer then was you can achieve the same result by specifying how many times you want to do it in a week and it will automatically handle it for you. While this does work in principal it creates annoying scenarios:
If you have a work habit that you want to do only during the week, if you do it successfully 5 days the weekend correctly gets implicit checks, but if you do it only 4 days, the weekend looks like failed days. We have the reminder active days to at least not remind you to do it then, but still visually it's not optimal.

This takes the same setting for reminders but goes one step further, on days that the habit is not active, it puts an automatic implicit checkmark. (P.S. in my `personal_dev` I put a `SKIPPED` mark so the calendar is shaded that these days are inactive, but in the current `dev` we interpret `SKIPPED` as an explicit action only).

### DIFF
--- a/android/uhabits-android/src/androidTest/java/org/isoron/uhabits/acceptance/steps/EditHabitSteps.java
+++ b/android/uhabits-android/src/androidTest/java/org/isoron/uhabits/acceptance/steps/EditHabitSteps.java
@@ -69,9 +69,9 @@ public class EditHabitSteps
         onView(withId(R.id.done_button)).perform(click());
     }
 
-    public static void clickReminderDays()
+    public static void clickActiveDays()
     {
-        onView(withId(R.id.reminderDatePicker)).perform(click());
+        onView(withId(R.id.activeDaysDatePicker)).perform(click());
     }
 
     public static void unselectAllDays()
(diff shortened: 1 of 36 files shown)


KristianTashkov:
Looks like SQL added the `rename column` syntax later and while it worked through the emulator, the SQL tests were failing. For now I'm adding a new column and leaving the old `reminder_days` but if we end up merging it we should probably do the whole ordeal with a backup table that you copy to/from which you later drop.

KristianTashkov:
Since you reverted the visual change to implicit checkmarks I made inactive days be skips here as well.

chennemann:
As mentioned in #630 I do like this feature since it would allow us to more specifically react to the State of a habit and calculate implicit days better

KristianTashkov:
Playing around with this version it has a bug where an implicit skip is an inescapable state by toggling (since it goes to `NO` which gets converted to `SKIP` again). This probably has to be solved by making a new `SKIP_AUTO` checkmark state that goes to checked when toggling.

iSoron:
Let's keep the discussion in #630.

PR #634, opened 2020-09-04
Update widgets at midnight

KristianTashkov:
Currently widgets don't get updated after midnight until you take an action on one of them.

### DIFF
--- a/android/uhabits-core/src/main/java/org/isoron/uhabits/core/utils/DateUtils.java
+++ b/android/uhabits-core/src/main/java/org/isoron/uhabits/core/utils/DateUtils.java
@@ -175,9 +175,14 @@ public abstract class DateUtils
         return getStartOfDay(getLocalTime());
     }
 
+    public static long getTomorrowStart()
+    {
+        return getUpcomingTimeInMillis(0, 0);
+    }
+
     public static long millisecondsUntilTomorrow()
     {
-        return getStartOfToday() + DAY_LENGTH - getLocalTime();
+        return getTomorrowStart() - getLocalTime();
     }
 
     public static GregorianCalendar getStartOfTodayCalendar()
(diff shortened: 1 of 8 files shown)


android/uhabits-core/src/main/java/org/isoron/uhabits/core/utils/DateUtils.java
```diff
@@ -175,9 +175,14 @@ public static long getStartOfToday()
         return getStartOfDay(getLocalTime());
     }
 
+    public static long getTomorrowStart()
```
iSoron, line 178: For consistency, please rename this to `getStartOfTomorrow()`

android/uhabits-core/src/test/java/org/isoron/uhabits/core/utils/DateUtilsTest.java
```diff
@@ -164,6 +164,7 @@ public void testTruncate_year()
     @Test
     public void testMillisecondsUntilTomorrow() throws Exception
     {
+        DateUtils.setFixedTimeZone(TimeZone.getTimeZone("GMT"));
```
iSoron, line 167: Was this required to make the tests pass in your computer? It makes sense, but I wonder why we haven't run into any issues before.

iSoron (changes requested):
Thanks for the pull request, @KristianTashkov. It looks good to me. I had only some minor comments.

android/uhabits-core/src/main/java/org/isoron/uhabits/core/utils/DateUtils.java
KristianTashkov: done!

android/uhabits-core/src/test/java/org/isoron/uhabits/core/utils/DateUtilsTest.java
KristianTashkov: because after this change `MillisecondsUntilTomorrow` goes through  `getUpcomingTimeInMillis` which is time zone dependent

iSoron (approved):

PR #638, opened 2020-09-12
Fix clearing of reminders

KristianTashkov:
Fixes clearing a reminder that is currently broken because save only touches the reminder if it's not off.


iSoron:
Thank you for the pull request. I think we could have some automated UI tests to make sure that reminders are saved correctly. We currently have UI tests for frequency and other fields.

PR #639, opened 2020-09-12
Don't show reminders from archived habits

KristianTashkov:
Currently when you archive a habit with a pending reminder, it still gets shown that day. Afterwards it correctly doesn't schedule it anymore.

### DIFF
--- a/android/uhabits-core/src/main/java/org/isoron/uhabits/core/ui/NotificationTray.java
+++ b/android/uhabits-core/src/main/java/org/isoron/uhabits/core/ui/NotificationTray.java
@@ -194,18 +194,27 @@ public class NotificationTray
                 return;
             }
 
-            if (!shouldShowReminderToday()) {
-                systemTray.log(String.format(
-                        Locale.US,
-                        "Habit %d not supposed to run today. Skipping.",
-                        habit.id));
-                return;
-            }
-
             if (!habit.hasReminder()) {
                 systemTray.log(String.format(
                         Locale.US,
                         "Habit %d does not have a reminder. Skipping.",
+                        habit.id));
+                return;
+            }
+
+            if (habit.isArchived())
+            {
+                systemTray.log(String.format(
+                        Locale.US,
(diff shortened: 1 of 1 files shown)


android/uhabits-core/src/main/java/org/isoron/uhabits/core/ui/NotificationTray.java
```diff
@@ -202,23 +202,16 @@ public void onPostExecute()
...
             }
 
-            if (!habit.hasReminder()) {
-                systemTray.log(String.format(
-                        Locale.US,
-                        "Habit %d does not have a reminder. Skipping.",
-                        habit.id));
-                return;
-            }
-
```
iSoron, line 212: Could you move this block up (before the `shouldShowReminderToday` block) instead of removing it?

android/uhabits-core/src/main/java/org/isoron/uhabits/core/ui/NotificationTray.java
```diff
@@ -202,23 +202,16 @@ public void onPostExecute()
...
             systemTray.showNotification(habit, getNotificationId(habit), timestamp,
                 reminderTime);
         }
 
         private boolean shouldShowReminderToday()
         {
             if (!habit.hasReminder()) return false;
-            Reminder reminder = habit.getReminder();
+            if (habit.isArchived()) return false;
```
iSoron, line 212: Instead of modifying this method, could you add a new block in `onPostExecute` instead, where we print `"habit is archived, skipping"`? I would like the log to tell us **precisely** why have we decided not to show a notification. This helps us debug "notifications are not showing up" bug reports.

iSoron (changes requested):
Thank you for the pull request, @KristianTashkov

iSoron (changes requested):
Thank you for the pull request, @KristianTashkov

KristianTashkov:
@iSoron Done :)

iSoron:
Awesome, thanks!

PR #643, opened 2020-09-14
Feat/decouple checkmark state and source

chennemann:
So I took the time and split the checkmark state from the input source.
So now instead of the states YES_MANUAL and YES_AUTO we only have YES plus a boolean flag indicating whether it was a manual Input.

Locally I still get two failed tests but I think this has something to do with my local setup and not with the changes from this PR.
I would appreciate if someone could verify this though.

The HabitCSVExporterTest > testExportCSV fails because the created tmp dir can not be deleted
The ListHabitsBehaviorTest > testOnExportCSV_fail fails because apparently the 'setWritable(false)' method call does not have the desired effect.

Besides that I hope I followed the guidelines and this can be merged. Feedback is appreciated.


android/uhabits-android/src/main/java/org/isoron/uhabits/widgets/views/CheckmarkWidgetView.java
```diff
@@ -49,74 +53,76 @@
...
-
-            case Checkmark.YES_AUTO:
-            case Checkmark.NO:
-            default:
-                getResources().getString(R.string.fa_times);
-                bgColor = res.getColor(R.attr.cardBgColor);
-                fgColor = res.getColor(R.attr.mediumContrastTextColor);
-                setShadowAlpha(0x00);
-                break;
+        if (checkmarkState.isManualInput()) {
```
chennemann, line 83: I'm not 100% sure whether this is the desired behaviour. So this needs to be clarified.

iSoron:
Thank you for the PR, @KristianTashkov. Could you please explain what benefits do you believe this PR brings to our code base? It seems like all these changes (including changes to our database schema) were made so that you could replace:

> r.getValue() != Checkmark.YES_MANUAL

by 

> r.getValue() == Checkmark.YES && r.isManualInput()

In my opinion, this is actually a regression; the first is more concise and easier to read.

Are there any other places where these changes make the code simpler?

iSoron:
Also, just a quick note, but everything we store in the database is user-provided. Everything we compute is only kept in memory and it's recomputed every time we restart the app.

chennemann:
I see your point and I agree that at the current state we don't gain too much value from this.
I see two advantages here:

1. It's more explicit -> Easier to reason about
2. We had multiple discussions about different features that would require the addition of additional states like NO_AUTO/NO_MANUAL, SKIP_AUTO/SKIP_MANUAL, UNKNOWN. Your argument in those cases was always that we should try to minimize the available states because they make it hard to reason about the state of the repetition. And that point is very valid but the underlying issue is the encoding of the source (manual/auto) into the state. So this PR tries to lay the foundation for those ideas that require the knowledge about both properties while simultaneously keeping the states to a minimum level.

I'll look into the failed check when we decided to go forward with this PR.

iSoron:
Thank you for the pull request again, @chennemann, but as mentioned before, I don't think the benefits to our code base are large enough to justify such major refactoring.

PR #646, opened 2020-10-02
Show score and the current streak as text

asdoi:
I created this small UI change for my personal use but maybe this could also be useful for other users:
I personally do not like the representation of the score with a circle because on my phone the circle is very small and I do not get much information through it.
Furthermore I want to see my current streak in the main view and do not want switch to the details page every time I just want to know how many days I now succeeded.

This adds a setting to display the score as text.
If enabled the circle is replaced by a TextView that displays the percentage and if the habit is checked for today or yesterday the current streak is displayed next to the score.
Moreover if the current streak is the highest streak it is displayed big with the habit color if not it is just grey and small.

Screen


rahulps108:
beautiful

asdoi:
Well, yes, I could increase the font size in the code but then the score is bigger than the name of the habit, which I think does not look so good.

iSoron:
Thank you for the pull request, @asdoi, but the visual representation seems to be working well for most users; this is not something I have heard complaints about.

Issue #545, opened 2020-01-05
Use dark theme also in settings window


szpak:
Having the dark theme enabled I was surprised to open the settings page and be glared by the while background there. It's not a big issue, but it the back theme could be used consistently.

1.8.5 on Android 10.

iSoron:
Thank you for the feedback, @szpak. I agree that the settings should follow dark theme too. Pull requests are welcome.

ayseenver:
Hi both, 

I was going to try and fix this but on my version of the app (1.7.11), the settings page is dark on night mode. Does it still need fixing?

szpak:
@ayseenver Just tried with 1.8.7 and the settings page is still white. The dark theme ticked in the app.

MRoyDroid:
Hi, @iSoron I've implemented the Dark Theme on Settings Screen. I've created a new Dagger Component (SettingsActivityComponent) for SettingsActivity to get access to the ThemeSwitcher. But I have a confusion, as I've used the HabitsApplicationComponent to build my new SettingsActivityComponent. Is that okay or do I need to create a SettingsApplicationComponent for that?

Sunxy88:
Hi @iSoron We are a group of students trying to contribute on this issue. Since this issues is still open, we suppose we could have a try. However, we are all beginners in android development. We will try our best to do so in a month and it could not be more fantastic if you give us some advices to contribute in more efficient way and to keep codes as elegant as yours ? If we finish this issue sooner than we thought, we would try to solve other issues.
Thank you very much!

iSoron:
Hi @Sunxy88, this issue is still open and pull requests are very welcome. If you would like, please feel free to submit an early version (work-in-progress) of your proposed changes, and we can start a discussion about the implementation.

PR #655, opened 2020-11-11
Use dark theme in settings window.

Sunxy88:
#545  Add 4 lines of code referred to [EditHabitActivity.kt](android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/edit/EditHabitActivity.kt) to apply dark theme in settings window.

### DIFF
--- a/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/settings/SettingsActivity.java
+++ b/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/settings/SettingsActivity.java
@@ -23,17 +23,27 @@ import android.os.*;
 
 import org.isoron.androidbase.activities.*;
 import org.isoron.androidbase.utils.*;
+import org.isoron.uhabits.HabitsApplication;
+import org.isoron.uhabits.HabitsApplicationComponent;
 import org.isoron.uhabits.R;
+import org.isoron.uhabits.activities.AndroidThemeSwitcher;
 
 /**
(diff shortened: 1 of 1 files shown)


android/uhabits-android/src/main/java/org/isoron/uhabits/activities/settings/SettingsActivity.java
```diff
@@ -23,17 +23,27 @@
 
 import org.isoron.androidbase.activities.*;
 import org.isoron.androidbase.utils.*;
+import org.isoron.uhabits.HabitsApplication;
```
iSoron, line 26: Please use wildcard imports

iSoron (changes requested):
Thank you for the pull request, @Sunxy88. I have tested it and it works great. I only have one comment about code style. After that is fixed, I can merge it.

iSoron (approved):

iSoron:
Looks great now. Thanks again, @Sunxy88!

szpak:
Unfortunately, it doesn't work for me with 1.8.9 and 1.8.10 from Google Play. Dark theme enabled, but the settings panel is as white as it was before. I forced the app to restart. Android 11.

iSoron:
@szpak This fix will become available in Loop 2.0

szpak:
@iSoron Thanks for your explanation, I've just taken a brief look into default branch (dev), not spotting that 1.8.x is from master.

Issue #579, opened 2020-05-03
Highlight habits that need to be done today


kathytu:
Hello! I'm new to Loop and I love it so far. Thanks to everyone for working on this app.

I have a question about easily seeing what still needs to be accomplished for today. I know currently you see a light grey "x" for undone habits, and as well I know you can choose to hide completed habits. I'd love to be able to see all my habits at once, but just the ones that need to be done are somehow highlighted or emphasized. 

Suggestions on how this could be implemented:

- the entire row of the habit can be highlighted
- instead of a "x" maybe a more noticeable checkbox
- make the "x" a darker shade of grey

Thanks so much!

chennemann:
Thanks for your suggestions. I agree with you that it's not visible all that well. I like your third suggestion to improve the contrast for a better readability. 
We need to see however what side effects a change like that would have.

iSoron:
Mocks or prototypes showing alternative color schemes are very welcome, but I think it would be hard to make the "x" more noticeable while keeping the interface visually attractive.

Another idea could be showing the habits that need to be done at the top of the list, and the habits that have already been completed at the bottom. That is, we could implement a new sorting method. This should be fairly easy.

Vncntmrchl:
Hello!
My teammates and I are working on this issue. We are students learning Android development so our skills are still limited.
We managed to add a sorting option to show habits completed today at the bottom of the list. We were initially trying to define some kind of "priority ranking" among habits but it turned out a bit too complex.
Moreover, we'd like to make habits completed today a bit less opaque to make it even more obvious that they were already done. We are currently working on this feature but we are having a hard time finding a way to do so (where to change opacity for example).

iSoron:
> We managed to add a sorting option to show habits completed today at the bottom of the list.

Thanks for working on this @Vncntmrchl. I'm looking forward for the pull request.

> We were initially trying to define some kind of "priority ranking" among habits but it turned out a bit too complex.

Could you elaborate on that?

> Moreover, we'd like to make habits completed today a bit less opaque to make it even more obvious that they were already done.

For such visual changes in the user interface, I would recommend creating a mockup first (for example, using GIMP or Photoshop), before spending too much time on the actual Java/Kotlin implementation. I am concerned that making the card less opaque will make it too hard to read.

Vncntmrchl:
Hello again!

I just realized that @hiqua did a really great job at answering this issue. What we did is actually included in his own pull request so I don't think it is necessary for me to do another pull request.
[...]

Issue #565, opened 2020-03-23
Hide completed, but only if you've explicitly checked them today


akfaew:
I track around 25 habits with the app, and sometimes struggle with organizing them.

I have some physiotherapy exercises that I intend to do 5 times in 7 days at a minimum to have peace of mind, but of course it's better to do them 7 times in 7 days.

I'd also like to shave at least once every three days, but naturally it's better to do it every day.

Currently there is a very convenient option: "Hide completed". It would, however, be more convenient to only hide items I've explicitly completed, showing all others, including those with an "automatic done".

And thank you for making the app @iSoron, it's been remarkably helpful.

NellyWhadsDev:
The current work flow is based on "at least N days in M days" to consider a habit "complete". The proposal seems to be to modify this to be "at least N days up to M days in P days".

This would fit a user story along the lines of "I'd like to shave at least 3 times a week, but 5 times is when I consider my habit schedule met."

This would however introduce another degree of freedom to the "completion" metric. The above example, with 4 days of shaving, would be a "partially complete" habit, since the hard target is 5 days, and the soft target is 3 days. From what I've seen so far in code-land, neither the UI or the score calculation is quite ready for this.

As cop-out-ish as it seems, the best way to accomplish this right now would probably be with 2 separate habits :(

hiqua:
I interpreted the proposal differently, and indeed, I was looking for this issue before creating my own describing what I think @akfaew tried to describe.

Basically there's no completion logic change, but only a new UI option, which is to add a 3rd option to the already existing:

1) Hide archived
2) Hide completed

to end up with something like:

1) Hide archived
2) Hide completed
3) Hide ticked today

(note that here, 3) implies 2) so it's not ideal).

When you have option 3) activated, you would only see the activities you can still tick.


By default, every day, I just want to see what's left to tick. I have a preference for uncompleted habits of course, but if I can tick completed habits it's also good.

To summarize, it's just one more UI filter that does not collide with any user that would not use it. I think there's still the question of how to make 2) and 3) work together nicely though.

akfaew:
Yes, that's what I meant. Sorry if I didn't write it clearly enough. Thank you @hiqua.

kkelleyjr:
I think this is a much needed feature and would be massively helpful. I think this feature would also have to include an additional completion option, x for not done, ✔️ for done, and / for intentionally not doing. With this adjustment we are able to get to a empty screen for 💯 for the day.

[...]

PR #660, opened 2020-11-15
Allow user to sort by status

hiqua:
Allow user to sort by priority, i.e. in this order
* not done (NO)
* "done" because done the previous days and the habit requires less than once per day (YES_AUTO), but not ticked today
* done (YES_MANUAL)

Partially solves #565 and #579.

[Screenshot_20201115_153752](https://user-images.githubusercontent.com/3812042/99187747-a42dc380-2758-11eb-88f7-2212fd96cbed.png)

### DIFF
--- a/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/HabitCardListAdapter.java
+++ b/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/HabitCardListAdapter.java
@@ -83,6 +83,7 @@ public class HabitCardListAdapter
         cache.setListener(this);
         cache.setCheckmarkCount(
             ListHabitsRootViewKt.MAX_CHECKMARK_COUNT);
+        cache.setPreviousOrder(preferences.getDefaultPreviousOrder());
         cache.setOrder(preferences.getDefaultOrder());
 
         setHasStableIds(true);
@@ -329,11 +330,6 @@ public class HabitCardListAdapter
     public HabitList.Order getOrder()
     {
         return cache.getOrder();
-    }
-
-    @Override
-    public HabitList.Order getPreviousOrder() {
-        return cache.getPreviousOrder();
     }
 
     /**
--- a/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/HabitCardListAdapter.java
+++ b/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/HabitCardListAdapter.java
@@ -320,9 +320,20 @@ public class HabitCardListAdapter
     }
 
     @Override
+    public void setPreviousOrder(HabitList.Order order) {
+        cache.setPreviousOrder(order);
+        preferences.setDefaultPreviousOrder(order);
+    }
+
+    @Override
     public HabitList.Order getOrder()
     {
         return cache.getOrder();
+    }
+
+    @Override
+    public HabitList.Order getPreviousOrder() {
+        return cache.getPreviousOrder();
     }
 
     /**
--- a/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/HabitCardListAdapter.java
+++ b/android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/HabitCardListAdapter.java
@@ -21,7 +21,7 @@ package org.isoron.uhabits.activities.habits.list.views;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
-import androidx.appcompat.widget.*;
+
 import android.view.*;
 
 import androidx.recyclerview.widget.RecyclerView;
@@ -83,8 +83,8 @@ public class HabitCardListAdapter
         cache.setListener(this);
         cache.setCheckmarkCount(
             ListHabitsRootViewKt.MAX_CHECKMARK_COUNT);
-        cache.setPreviousOrder(preferences.getDefaultPreviousOrder());
-        cache.setOrder(preferences.getDefaultOrder());
+        cache.setSecondaryOrder(preferences.getDefaultSecondaryOrder());
+        cache.setPrimaryOrder(preferences.getDefaultPrimaryOrder());
 
         setHasStableIds(true);
     }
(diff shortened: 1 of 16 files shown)


iSoron:
Thank you for the pull request, @hiqua. After testing this, I think it would actually make more sense for us to add a checkbox "Move completed habits to the bottom". In this way, the user could still sort by score, for example, but have the items move out of the way when completed. I think it would be more complicated to implement than your current solution, but much better from a user perspective.

Would you like to work on this?

hiqua:
@iSoron thanks for the review. Another possibility would be to remember the previous sorting method along with the current one, and if the current one returns an equality, compare with the previous one. To have the configuration you mention, a user would then need to:
* select score sorting
* select status sorting

From the user interface, we could just keep things like they currently are. The feature would thus be a bit hidden, but still discoverable enough.

If you don't like this suggestion I'll implement yours on a new PR.

iSoron:
> Another possibility would be to remember the previous sorting method along with the current one, and if the current one returns an equality, compare with the previous one.

I like this solution. It would also work with other pairs of sorting methods (e.g. "color" then "score").

hiqua:
Sorting by color then priority:
Screenshot_20201117_232541

Sorting by name then priority:
Screenshot_20201117_232604

iSoron:
Thanks for the updates, @hiqua. A few comments:

1. Could you rename "sort by priority" to "sort by status"? Status sounds more clear to me. "Priority" sounds like something that the user could assign to each habit.
2. It seems like we are not persisting the secondary sorting method, only the primary one, so the ordering changes after the app restarts.
3. Most comparators we previously had (color, score, position) are hardcoded to fall back to comparison by name in case of ties. I think it makes sense to let them fall back to the secondary sorting method instead. This would allow some interesting things such as sorting by color, but moving completed habits to the bottom of the color category when completed ("sort by status" then "sort by color").

hiqua:
1. done
2. done (I think), but I'm gonna need a review for that, it was more code than I expected. I should probably write more tests as well, but I'll wait for you to tell me whether you're somewhat satisfied with the current implementation first 
3. done

android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/HabitCardListAdapter.java
```diff
@@ -83,6 +83,7 @@ public HabitCardListAdapter(@NonNull HabitCardListCache cache,
         cache.setListener(this);
         cache.setCheckmarkCount(
             ListHabitsRootViewKt.MAX_CHECKMARK_COUNT);
+        cache.setPreviousOrder(preferences.getDefaultPreviousOrder());
         cache.setOrder(preferences.getDefaultOrder());
```
iSoron, line 87: Could you please rename `order` and `previousOrder` to `primaryOrder` and `secondaryOrder` everywhere? I think this is more clear. For example, these two call would become `cache.setPrimaryOrder(...)` and `cache.setSecondaryOrder(...)`

android/uhabits-core/src/test/java/org/isoron/uhabits/core/io/HabitsCSVExporterTest.java
```diff
@@ -68,11 +68,11 @@ public void testExportCSV() throws IOException
...
-        assertPathExists("001 Wake up early");
-        assertPathExists("001 Wake up early/Checkmarks.csv");
-        assertPathExists("001 Wake up early/Scores.csv");
-        assertPathExists("002 Meditate/Checkmarks.csv");
-        assertPathExists("002 Meditate/Scores.csv");
+        assertPathExists("001 Meditate/Checkmarks.csv");
+        assertPathExists("001 Meditate/Scores.csv");
+        assertPathExists("002 Wake up early");
+        assertPathExists("002 Wake up early/Checkmarks.csv");
+        assertPathExists("002 Wake up early/Scores.csv");
```
iSoron, line 75: Why has this changed? It seems unrelated to the feature.

android/uhabits-core/src/test/java/org/isoron/uhabits/core/models/HabitListTest.java
```diff
@@ -131,56 +130,73 @@ public void testOrdering()
...
         h4.setColor(1);
         h4.setPosition(2);
+        HabitList list;
 
-        list.add(h3);
-        list.add(h1);
-        list.add(h4);
-        list.add(h2);
-
+        list = testOrderingSetup(h1, h2, h3, h4);
```
iSoron, line 135: Why do we need to recreate the list repeatedly? Does the previous code not work anymore?

android/uhabits-core/src/test/java/org/isoron/uhabits/core/io/ImportTest.java
```diff
@@ -81,7 +81,7 @@ public void testRewireDB() throws IOException
 
         assertThat(habitList.size(), equalTo(3));
 
-        Habit habit = habitList.getByPosition(0);
+        Habit habit = habitList.getByPosition(1);
```
iSoron, line 84: Same as the previous comment.

iSoron (changes requested):
Thank you for the changes. Please see some comments below.

Regarding tests, I would suggest better testing the logic within `onSortToggleBy`. For example, we could call `onSortByName` and `onSortByStatus` a few times and make sure that all the `setOrder` and `setPreviousOrder` call are correct.

android/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/HabitCardListAdapter.java
hiqua: Done.

android/uhabits-core/src/test/java/org/isoron/uhabits/core/io/HabitsCSVExporterTest.java
hiqua: The default secondaryOrder is by name: https://github.com/iSoron/uhabits/pull/660/files#diff-36c07e8fa292800a1e6f010c52b7d81b33e2f1bc3514577d0c60df3d50959a86R42

Before my changes, this test was using the position sort. Given that these habits have the same position (`0`), they were left as they are inserted, so sorted by id. The changes reflect the facts that the habits are now sorted by name.

android/uhabits-core/src/test/java/org/isoron/uhabits/core/io/ImportTest.java
hiqua: Also same answer, the default secondary ordering is by name, so the insertion ordering is no longer there.

hiqua:
Test implemented, nothing left to do from my side.

PR #699, opened 2021-01-02
Convert org.isoron.uhabits.core.utils to Kotlin

MarKco:
Converting to Kotlin the whole org.jsoron.uhabits.core.utils
package. First is DateFormats, others will come.


MarKco:
For each commit I test
* building
* the whole test suite in the project for the core project
* executing (and playing a bit with the app)

is this enough?

iSoron:
Thank you for the pull request, @MarKco. GitHub actions runs all JVM tests (`./build.sh build`) and medium tests (`./build.sh medium-tests`). These two should be enough.

MarKco:
My bad. I have some issues running medium-tests on my machine, I'll manage to make it work, fail locally and fix MidnightTimer accordingly.

MarKco:
I managed to make the _medium-tests_ command work on my machine but still I must admit I'm pretty stuck.

I see the error is the following:

```
Mockito cannot mock/spy because :
 - final or anonymous class
	at org.isoron.uhabits.activities.habits.list.views.HeaderViewTest.setUp(HeaderViewTest.java:53)
```

as far as I know this depends on Kotlin having final classes by default. I searched a bit and found that mockito > 2.10 should solve the issue. Also I tried by adding the org.mockito.plugins.MockMaker file with the "mock-maker-inline" content, [as suggested by Antonio Leiva](https://antonioleiva.com/mockito-2-kotlin/), but still didn't get any result. I added it in both modules, uhabits-android and uhabits-core (since HeaderViewTest is in android module while MidnightTimer is in the core), still nothing changes. I must admit I didn't really expect the fix to work because I already see the following line in the build.gradle, so I guess inline mocking is already enabled module-wide:

    testImplementation "org.mockito:mockito-inline:2.13.0"

I even tried - for testing purposes - to make the MidnightTimer class open

```
@AppScope
open class MidnightTimer @Inject constructor() {
    private val listeners: MutableList<MidnightListener> = LinkedList()
```

This should have fixed the issue but actually it didn't. I'm pretty surprised. I should probably study deeply how mockito works in order to understand this better.

Do you recall if it ever happen before during the Java-Kotlin conversion? I checked in the issues, couldn't find anything related.

Thank you

MarKco:
Partial amend: I now see that `./build.sh medium-tests` doesn't rebuild the package, so I must perform a build every time before a new call to `./build.sh medium-tests`. 

I tried again, performing a full build/test flow, and
* I don't get the "final or anonymous class" error if I make the `MidnightTimer` class open
* I still get the error if I don't open the class. Therefore the "inline" tricks on mockito are not working as expected.

iSoron:
Hi Marco, feel free to make the classes open for now. If we figure out how to make mockito work with closed Kotlin classes, we can make them closed again in the future. Pull requests for that are also welcome.

To run both `build` and `medium-tests` in a single command, you can run `./build.sh build && ./build.sh medium-tests`. These steps are separate because GitHub Actions runs `build` once and `medium-tests` multiple times (one for each Android version, running on a different emulators).

MarKco:
The "ready for review" action was triggered by mistake due to my very sensitive touchpad. Sorry. I switched back to draft waiting for the whole utils package to be converted successfully.

MarKco:
I had errors performing medium-tests on my emulator but I think they were due to the different configuration on the emulator itself. I see all is good here on the github emulators. Now the whole utils package of the core module has been converted to Kotlin.

iSoron:
Thank you for the pull request, @MarKco. Everything looked good, so I went ahead and merged it.

PR #707, opened 2021-01-12
Kotlin conversions

hiqua:
Trying to stick to one commit == one module at most.

Also fixes the klint problem we discussed.


iSoron:
Thank you for the pull request, @hiqua!

Issue #694, opened 2020-12-30
Make delayed midnight an opt-in feature


iSoron:
## Description

Since #621 was merged, Loop now extends all days a few hours past midnight (until 3AM). This is probably unexpected behavior for most users. We should make this an opt-in feature instead.

## Implementation

Most of the required code is already implemented. We just have to add this option to the settings screen and call `DateUtils.setStartDayOffset` with the correct arguments when the application starts (in `HabitsApplication.onCreate`).

## Mockup

![Screenshot_1609357696](https://user-images.githubusercontent.com/4699657/103377923-5837a380-4aa6-11eb-8415-1fa3bca91011.png)

PR #717, opened 2021-01-22
Midnight delay option

hiqua:
#694

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/ButtonPanelView.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/ButtonPanelView.kt
@@ -25,6 +25,7 @@ import android.view.View.MeasureSpec.EXACTLY
 import android.widget.LinearLayout
 import org.isoron.uhabits.R
 import org.isoron.uhabits.core.preferences.Preferences
+import org.isoron.uhabits.core.utils.DateUtils.Companion.setStartDayOffset
 import org.isoron.uhabits.utils.dim
 import org.isoron.uhabits.utils.toMeasureSpec
 
@@ -54,6 +55,12 @@ abstract class ButtonPanelView<T : View>(
 
     @Synchronized
     protected fun inflateButtons() {
+        if (preferences.isMidnightDelayEnabled) {
+            setStartDayOffset(3, 0)
+        } else {
+            setStartDayOffset(0, 0)
+        }
+
         val reverse = preferences.isCheckmarkSequenceReversed
 
         buttons.clear()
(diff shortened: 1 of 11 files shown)


uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/ButtonPanelView.kt
```diff
@@ -54,6 +55,12 @@ abstract class ButtonPanelView<T : View>(
 
     @Synchronized
     protected fun inflateButtons() {
+        if (preferences.isMidnightDelayEnabled) {
+            setStartDayOffset(3, 0)
+        } else {
+            setStartDayOffset(0, 0)
+        }
+
```
iSoron, line 63: Is this required because the previous code only runs during startup? I think if we need this here, then we almost certainly need it in many other places (for example, in `HeaderView`). I guess this feature is more tricky than I expected. Perhaps we should force an app restart after the user changes this preference.

uhabits-android/src/main/java/org/isoron/uhabits/notifications/SnoozeDelayPickerActivity.kt
```diff
@@ -49,7 +49,8 @@ class SnoozeDelayPickerActivity : FragmentActivity(), OnItemClickListener {
         val app = applicationContext as HabitsApplication
         val appComponent = app.component
         reminderController = appComponent.reminderController
-        habit = appComponent.habitList.getById(ContentUris.parseId(intent.data))
+        val data = intent.data
+        if (data != null) habit = appComponent.habitList.getById(ContentUris.parseId(data))
```
iSoron, line 53: Please revert if this is not directly related to the feature.

uhabits-android/src/main/java/org/isoron/uhabits/receivers/ReminderReceiver.kt
```diff
@@ -46,7 +46,8 @@ class ReminderReceiver : BroadcastReceiver() {
         Log.i(TAG, String.format("Received intent: %s", intent.toString()))
         var habit: Habit? = null
         val today: Long = getStartOfTodayWithOffset()
-        if (intent.data != null) habit = habits.getById(ContentUris.parseId(intent.data))
+        val data = intent.data
+        if (data != null) habit = habits.getById(ContentUris.parseId(data))
```
iSoron, line 50: Please revert.

iSoron (changes requested):
Thanks for the pull request, @hiqua. I just realized that this feature may be more tricky than expected. See comment below. Please let me know your thoughts.

uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/ButtonPanelView.kt
hiqua: Ok then I think https://github.com/iSoron/uhabits/pull/739 might be more reasonable.

iSoron:
Merged #739 instead of this.

PR #718, opened 2021-01-22
Disable transparency for stacked widgets

hiqua:
I've added a "stacked" argument and changed the getter accordingly. Honestly I'm not sure you'll find it clean, so let me know if you'd rather have a different solution.


iSoron:
Looks good, thanks!

PR #739, opened 2021-01-27
Implement the midnight delay option (alternative)

### LINKED ISSUE #694: Make delayed midnight an opt-in feature

hiqua:
Closes #694.

Follow-up from https://github.com/iSoron/uhabits/pull/717: I think it actually makes sense to tell the user that the app will have to restart for the setting to take effect, and keep the implementation clean, given that users will probably not tick the setting on and off regularly anyway. That would be this PR.

Let me know if I should go back to the previous one and make sure the setting applies right away.

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/HabitsApplication.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/HabitsApplication.kt
@@ -24,7 +24,7 @@ import android.content.Context
 import org.isoron.uhabits.core.database.UnsupportedDatabaseVersionException
 import org.isoron.uhabits.core.reminders.ReminderScheduler
 import org.isoron.uhabits.core.ui.NotificationTray
-import org.isoron.uhabits.core.utils.DateUtils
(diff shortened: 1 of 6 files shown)


iSoron:
Thanks! I have merged it for now. We can improve it in the future.

The fundamental issue here is that many objects throughout the application are querying the current date by communicating directly with `DateUtils`. Ideally, I think every object that requires the current date should receive it from whoever instantiated it, and only the entry points (`activity.onCreate`) should have access to the actual system date. I have already started this refactoring a while ago by removing the concept of "today" from all classes in `uhabits.core.models`, but we still have a long way to go.

PR #752, opened 2021-01-28
Set dataOffset when attaching a new CardView

### LINKED ISSUE #713: Some Habits Swap Days Displayed When Time-Scrolled While Off-Screen
## Pre-submission checklist
- [ x] I am submitting a bug report, not a feature request.
- [ x] I am running the latest version of Loop Habit Tracker.
- [ x] I have have searched for similar issues, but did not find any matches.

## Description
Days temporarily get moved around if you scroll back in time while they are off screen then you bring them back on screen.

## Steps to reproduce
1. Go to main display of all habits
2. Scroll down so some habits are off screen. Turning into landscape mode can help you have enough habits to accomplish this.
3. Drag on the list of days to move back in time and reveal older days.
4. Scroll back up to reveal the habits that were off-screen. Some of them should be shifted a few days off.
5. To get it back to correct, simply slide the days again, they will fall back in place.

## System information
 - Phone: Pixel 3
 - Phone Operating System: Android 11
 - App version: Always newest version. I download releases on github immediately.

This can be confusing to someone, gving false information about a habit. I got quite confused until I figured it out.

hiqua:
Fixes #713.

### DIFF
--- a/uhabits-android/src/androidTest/java/org/isoron/uhabits/acceptance/steps/CommonSteps.kt
+++ b/uhabits-android/src/androidTest/java/org/isoron/uhabits/acceptance/steps/CommonSteps.kt
@@ -69,6 +69,10 @@ object CommonSteps : BaseUserInterfaceTest() {
         device.waitForIdle()
     }
 
+    fun offsetHeaders() {
+        device.swipe(750, 160, 600, 160, 20)
+    }
+
     fun scrollToText(text: String?) {
         try {
             if (device
--- a/uhabits-android/src/androidTest/java/org/isoron/uhabits/regression/HasButtonsViewAssertion.kt
+++ b/uhabits-android/src/androidTest/java/org/isoron/uhabits/regression/HasButtonsViewAssertion.kt
@@ -1,0 +1,24 @@
+package org.isoron.uhabits.regression
+
+import android.view.View
+import androidx.test.espresso.NoMatchingViewException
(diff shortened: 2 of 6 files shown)


iSoron:
Thank you for the pull request, @hiqua! This is a pretty serious bug, in my opinion, so I think we should have some tests to make sure it never comes back. Would you like to try writing an Espresso test to replicate this bug before we fix it? I can also work on this myself if you're busy.

hiqua:
Too bad the CI doesn't run the test, I wanted it to fail on the new test to show that the fix does fix the bug. I tested it locally and I think both the tests and the fix are correct though. Please let me know what I should still improve.

iSoron:
Some large tests were failing on `dev` before even applying your PR, so I went ahead and fixed them. I will review your PR soon.

In the meanwhile, if you are running the large tests manually (on Android Studio), instead of using `./build.sh large-tests`, please try the script. It creates a fresh AVD with fixed settings, so it's more reproducible.

hiqua:
> In the meanwhile, if you are running the large tests manually (on Android Studio), instead of using `./build.sh large-tests`, please try the script. It creates a fresh AVD with fixed settings, so it's more reproducible.

Ok thanks I'll do that in the future to test before I submit the PR.

iSoron:
Merged in 2dcce080d34b with some minor code cleanup in 5be2c51d7. Thanks again!

Issue #669, opened 2020-11-20
Improve list of translators


iSoron:
## Description

In the about screen, we have a list of everyone who helped translating Loop Habit Tracker into different languages. This list has gotten very long over time, and it's now quite disorganized. It's also missing [a lot of people who have contributed more recently](https://translate.loophabits.org/project/loop-habit-tracker/activity_stream).

### Subtasks

- [x]  To make it more organized, it would be nice if the list was grouped by language. We could also potentially make each language group collapsible, so it's easier to navigate.
- [x] Modifying the list currently requires one of the developers to modify [a long XML file](https://github.com/iSoron/uhabits/blob/dev/android/uhabits-android/src/main/res/layout/about.xml). It would be nice if this list came from a file that was easier to edit. Extra points if this file could be exported automatically from our translation platform somehow. (#759)

iSoron:
@hiqua I have updated the top post to indicate that this is partially completed. I will close this issue when we group translators by languages.

hiqua:
Would it be ok to just have one text file per language and include in the .xml accordingly, with subsections?

For the collapsible part it seems that ExpandableListView would be the way to go.

PR #759, opened 2021-02-02
Implement task to generate translator layout from a text file

hiqua:
See #669.

# TODO

* test on real project
* manually add translators that are not on crowdin

# DONE
* fetch list from API instead of using a local file (https://support.crowdin.com/api/v2/#operation/api.projects.members.getMany)
* generate layout XML
* add task dependency to some publish task so that it has to be run before a new APK is published (in build.sh)

# WONTDO (in this PR):
* collapsible groups


iSoron:
Nice usage of gradle and automatic code generation! Just a thought, but if gradle could detect that the file is already up-to-date (similar to what GNU Make does) and automatically skipped rewriting it, we could even run this task every time the project is compiled.

hiqua:
> Nice usage of gradle and automatic code generation! Just a thought, but if gradle could detect that the file is already up-to-date (similar to what GNU Make does) and automatically skipped rewriting it, we could even run this task every time the project is compiled.

Or even just run it every time, since it doesn't take much time anyway. But then it would mean we stick to this text file, instead of using the crowdin API. I thought the crowdin API was the preferred way, isn't it? That does mean that only a person with a crowdin token (configured somewhere for gradle to use) and the necessary rights (probably everyone but I have to check) can update the list.

What do you prefer? API or text file?

iSoron:
@hiqua Sorry, I just realized that you are planning to completely eliminate the `translators` file. Yes, that could work. In this case, as you said, we can run the gradle task (which fetches an updated list from crowdin and writes `about_translators.xml`) before publishing a new release, not on every build. One detail is that some translators contributed to the project before we started using crowdin, so they are not listed there. We will probably still need a `translators` file with these (not all) names.

> That does mean that only a person with a crowdin token (configured somewhere for gradle to use) and the necessary rights (probably everyone but I have to check) can update the list.

This should be public, but it's quite possible that they restrict API access to project administrators. Please let me know if crowdin doesn't allow you to fetch it.

hiqua:
I don't think they allow me to fetch the data, so you might have to do the final testing for the actual project (unless we change permissions).

I'm trying to use the crowdin library, but I think I encountered a bug, I'm waiting for an answer: https://github.com/crowdin/crowdin-api-client-java/issues/53

Once I'm able to get the permission for every project member it should work.

For the translators that are not on crowdin, I could just add another `include` in the layout. If we end up splitting by language and collapsible groups, we'll have to multiply these `include` declarations which is not ideal, but good enough IMHO.

iSoron:
@hiqua I have made you a manager of the Loop's crowdin project.

> How can I get the list of contributors that contributed before crowdin? Git history I guess?

I suggest just keeping the current list of translators, but removing the duplicates. We previously used another translation platform, and I manually added the translated files to git, so their names are not in git history either.

hiqua:
Ok thanks!

Apparently permissions are given to everyone, so that won't cut it to find who's contributed to what. I've asked the people at crowdin if there is a better way. The only solution I can think about now (which does not involve repeatedly hammering their servers) is to request one of their "reports" about the translations in excel format and extract the information from there. Needless to say, I'm not ecstatic about it.

So I'd suggest to just keep it to text files for now, until something better comes along.

hiqua:
Ready for review, but the commits should be rewritten / squashed before merge.

iSoron:
Merged! Thanks a lot, @hiqua.

Issue #719, opened 2021-01-22
Dark mode unchecked items have low contrast


melocene:
## Pre-submission checklist
- [x] I am submitting a bug report, not a feature request.
- [x] I am running the latest version of Loop Habit Tracker.
- [x] I have have searched for similar issues, but did not find any matches.

## Description
In the app when using dark mode the checkmarks/X for incomplete days has a low contrast and can be tough to see. Light mode is better in this regard as they use a grey on a light background. To me it seems the color is the same for the incomplete marks between light and dark mode, and the grey background of the task is too similar. Trying the AMOLED did not seem to change either color.

The color for tasks can be customized, and the completed marks use the same color, but I am also not seeing a way to customize other colors in the app.

## Steps to reproduce
1. Add a task
2. Enable dark mode
3. Observe the marks for incomplete activity

## System information
 - Phone: Pixel 4a
 - Phone Operating System: Android 11
 - App version: 1.8.11

hiqua:
Duplicate of https://github.com/iSoron/uhabits/discussions/413

iSoron:
Thank you for report @melocene, but this is expected app behavior, not a bug. Please feel free to upvote the feature request linked by @hiqua if you would like to see changes to the theme.

melocene:
I am not sure I agree with it being a feature request as this is an issue for those with light sensitivity, trouble seeing shades of grey, or are otherwise visually impaired and trying to use the app.

The idea behind this issue was to just suggest a small change to the icon contrast to make it more visible, not a request for an overhaul of the theme, or add additional settings.

I see the linked feature request, and actually found another in the discussions, showing there is need for this. Since you note it as intended, would it be possible to get more context as to why this is designed as it is vs being more accessible for the visually impaired users?

Issue #649, opened 2020-10-17
Dark Mode Cross Icon Visibility


shockergit:
Dev can you change the cross icon to red color in dark mode. it not visible

PR #777, opened 2021-02-07
Update X contrast color to medium to improve readability.

ErinMyLungs:
This is a PR to handle issue mentioned in 
* [issue #719](https://github.com/iSoron/uhabits/issues/719)
* [issue #649](https://github.com/iSoron/uhabits/discussions/649)
* [Issue #413](https://github.com/iSoron/uhabits/discussions/413)

I debated a few solutions like a ternary operator to up contrast on dark mode only but this solves the issue and while I don't use light mode frequently, the improved contrast is nicer there too.

### Light Mode Example
image
### Dark Mode Example
image

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/CheckmarkButtonView.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/CheckmarkButtonView.kt
@@ -116,7 +116,7 @@ class CheckmarkButtonView(
 
     private inner class Drawer {
         private val rect = RectF()
-        private val lowContrastColor = sres.getColor(R.attr.lowContrastTextColor)
+        private val medContrastColor = sres.getColor(R.attr.mediumContrastTextColor)
 
         private val paint = TextPaint().apply {
             typeface = getFontAwesome()
@@ -129,7 +129,7 @@ class CheckmarkButtonView(
             paint.color = when (value) {
                 YES_MANUAL -> color
                 SKIP -> color
-                else -> lowContrastColor
+                else -> medContrastColor
             }
             val id = when (value) {
                 SKIP -> R.string.fa_skipped


hiqua:
@ErinMyLungs if you could provide screenshots of the changes that could make it faster to review for @iSoron !

ErinMyLungs:
@hiqua Sorry, totally spaced that I should include images. Updated the main comment to include examples.

hiqua:
Apparently we're gonna have to update the screenshots of the test with this change. According to https://github.com/iSoron/uhabits/blob/dev/docs/TEST.md#running-medium-tests it should be enough to just run a few commands, if you have a proper testing setup. Do you want to do it?

ErinMyLungs:
> Apparently we're gonna have to update the screenshots of the tests with this change. According to https://github.com/iSoron/uhabits/blob/dev/docs/TEST.md#running-medium-tests it should be enough to just run a few commands, if you have a proper testing setup. Do you want to do it?

I was able to successfully run the medium tests which failed the same set of tests again but  running `./build.sh fetch-images` or `./build.sh accept-images` both throw errors saying the command doesn't exist.

I did some searching to try and find an alternate file that has that command and checked the git history to see if it used to be a command in `build.sh` but I came up empty.

If you have advice of where I should look, I'm down to try and run it again. Is this something that is defined in the repo somewhere or is it a standard command involved with the tooling that I don't have properly setup?

iSoron:
Thank you for the pull request, @ErinMyLungs, and welcome to the project!

To clarify, I think the main problem here is that the app currently renders `YES_AUTO`, `NO` and `UNKNOWN` using the same (low contrast) grey color, which makes it harder to distinguish between these three states. I agree that this is a problem, but I think increasing the contrast for all three states is not very aesthetically pleasing; it's also a major visual change for most users.

I would prefer the following less intrusive change:

- If `prefs.areQuestionMarksEnabled()` is false, then render `YES_AUTO` as medium-contrast grey and `NO` as low-contrast grey.
- If `prefs.areQuestionMarksEnabled()` is true, then render `YES_AUTO` and `NO` as medium-contrast grey, and `UNKNOWN` as low-contrast grey.

Regarding `fetch-images` and `accept-images`, I have restored these function in 589fd2f. Please rebase your branch, and you should be able to use them.

iSoron:
Closing due to inactivity. If you are still working on this PR, please let me know and I will reopen it.

grillatom:
hey @iSoron why you set people up to fail? i come to do similar pr and see this first.  you know damn well what your change does. your change still make X and ? not clear. the change screenshot look fine and since it for v2 https://semver.org/ says big change is ok due to major version.

you say "someone do pr" then they do and you say "no". if you intend to exclude users say so and accept what happens. dont carrot on stick people and play game with fake promise or imply things

hiqua:
@grillatom this is an open discussion, @iSoron didn't demand anything, so I'm not sure how you come to this kind of conclusion. The main problem with this PR was that the tests were not passing, which is obviously a blocker to get it merged.

Now regarding the solution itself, maybe you're right that it's better to have everything in higher contrast, but it'd be easier to say with screenshots comparing the two different implementations. You can do it and try to convince others using technical arguments, listing pros and cons.

avocadosec:
These accessibility options would be excellent to have as it is hard to utilize otherwise. I will have to continue using another habit tracking app until accessibility options such as this are implemented. Thanks!

qbarron73:
> 
> 
> hey @iSoron why you set people up to fail? i come to do similar pr and see this first. you know damn well what your change does. your change still make X and ? not clear. the change screenshot look fine and since it for v2 https://semver.org/ says big change is ok due to major version.
> 
> you say "someone do pr" then they do and you say "no". if you intend to exclude users say so and accept what happens. dont carrot on stick people and play game with fake promise or imply things

i agree with this comment, not sure why its hidden
here is proof the asked change dont help and also issue is in light mode. see X and ? still low visibility

screenshot_2021-03-11_06-53-08
screenshot_2021-03-11_06-52-03
screenshot_2021-03-11_06-52-28
screenshot_2021-03-11_06-52-56

iSoron:
Locking to prevent further spam from 1-day-old GitHub accounts. Please keep all discussion related to this feature restricted to #413. New PRs implementing the requested changes are still welcome.

PR #778, opened 2021-02-08
Fix shellcheck warnings

hiqua:
See https://github.com/koalaman/shellcheck

This commit fixes all the warnings while ignoring a few that can't be fixed.


iSoron:
Looks good, thanks!

PR #782, opened 2021-02-09
Add large tests workflow

hiqua:


iSoron:
I see that this workflow is still failing in [hiqua/uhabits2](https://github.com/hiqua/uhabits2/actions). Is the one in this PR different from that one?

hiqua:
No it's the same one. I think we'll have to adapt build.sh somehow if we want it to run correctly.

iSoron:
Implemented in b561dfe90d87caa74c7fe91af53caa79501b6016.

PR #783, opened 2021-02-09
Show "you're all done" when all habits are completed

hiqua:
I had to upgrade font awesome since the suggested character was not available in the current version.

Note that if all habits are archived, the message shown will be "Done for today" as if the user had completed some habits. I think that's fine, let me know otherwise.

Requires #784.


iSoron:
Thank you for the pull request, @hiqua! The code looks good, but I think at some point you removed the commit that updated the FontAwesome files, so I am getting a square character.

Screenshot_1613095494

hiqua:
Yes sorry, since the fontawesome update required a change I thought you'd prefer to have it as a separate PR.

iSoron:
Merged in 50ebff3edf48. I have also added view tests for `EmptyListView` in 11d55bbf02 and fixed a small bug in c53997ffc.

PR #784, opened 2021-02-10
Upgrade fontawesome to 5.15.2

hiqua:
Needed for the umbrella beach character.

I think they changed the bell character, I used this: https://fontawesome.com/icons/bell?style=solid


hiqua:
Still need to fix this

new
old

iSoron:
@hiqua Thank you for the pull request. It looks like the TTF file you have is for the `solid` variant. Could you switch to the `regular` variant? The icons are more similar to the ones we were previously using.

Also, did you download the TTF file from fontawesome.com, from a 3rd party website, or did you convert it yourself? I'm asking because I was trying to find the original TTF files on their website, but it seems like they now only offer SVG and OTF.

hiqua:
> Also, did you download the TTF file from fontawesome.com, from a 3rd party website, or did you convert it yourself? I'm asking because I was trying to find the original TTF files on their website, but it seems like they now only offer SVG and OTF.

https://fontawesome.com/how-to-use/on-the-web/setup/hosting-font-awesome-yourself `Download Font Awesome Free
for the Web ` and look for .ttf in the zip archive. I agree it's quite confusing...

hiqua:
@iSoron I took the `solid` variant because the `regular` one was not working (the icons were not rendered properly). I think it's because of the license, e.g. https://fontawesome.com/icons/times?style=regular has a `pro` label, meaning that we're not supposed to use it without a license.

So I see these options:
* keep both older version and new version of fontawesome
* switch everything to solid
* use a character already in FA 4 for the other PR so that we don't need FA 5

iSoron:
OK, I was not aware that some regular icons were pro-only ([some others are free](https://fontawesome.com/icons/bell?style=regular)). In this case, we can switch to solid.

PR #837, opened 2021-03-09
Make validation errors readable

### LINKED ISSUE #757: Validation errors in dark theme are unreadable
## Description
When the user tries to save a habit, but some data is missing, the app shows validation errors with a brief description of the problem. In the dark theme, the error messages are shown as black text on black background, which makes them completely unreadable. We should make the text white on black instead.

## Steps to reproduce
1. Open the app
2. Switch to the dark theme
3. Click "add new habit"
4. Click "save" without typing any data

## Screenshots

Screenshot_1612114857

## System information
 - Phone: Emulator
 - Phone Operating System: Android 10
 - App version: e4121b2

hiqua:
Fixes #757.

I felt that was not the way, which is why I thought I'd only give a few hints there: https://github.com/iSoron/uhabits/issues/757#issuecomment-779245103

But revisiting this now I'm under the impression that the only proper way to do things would be to migrate to TextInputLayout and follow the guide there: https://material.io/components/text-fields/android#outlined-text-field

Maybe you'll find this trivial and quick to do, but I didn't, hence this quick-and-dirty fix.

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/edit/EditHabitActivity.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/edit/EditHabitActivity.kt
@@ -23,15 +23,22 @@ import android.annotation.SuppressLint
 import android.content.res.ColorStateList
 import android.graphics.Color
 import android.os.Bundle
+import android.text.Html
+import android.text.Spanned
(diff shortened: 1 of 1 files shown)


iSoron:
Thank you for the PR! I agree it's not the most elegant solution, but it does fix the problem. We can replace it by a better solution later.

PR #840, opened 2021-03-12
Start snooze activity after later is selected from lock screen

### LINKED ISSUE #787: No response from notifications actions on lockscreen
## Pre-submission checklist
- [x] I am running the latest version of Loop Habit Tracker.
- [x] I have have searched for similar issues, but did not find any matches.

## Description
I want to snooze the notifications on the lock screen, but there is no response when I click it. It does not snooze and dismiss the notifications or direct me to unlock to complete the action.

## System information
 - Phone: Sony Xperia 5
 - Phone Operating System: Android 10
 - App version: 1.8.12

hiqua:
Fixes #787.

This seems to fix it, but honestly I'm not aware of the full implications, and I haven't tested it enough to be confident that it doesn't introduce other bugs. So help appreciated, otherwise if it points you in the right direction for a proper fix that's good enough for me!

The change in the manifest might or might not be needed (I have tested only with it), but I think the doc says it should be like this:
https://developer.android.com/training/notify-user/navigation#ExtendedNotification

Leaving it as a draft for now (since I don't think it's ready to merge as-is), but I would likely need help to make further progress.

### DIFF
--- a/uhabits-android/src/main/AndroidManifest.xml
+++ b/uhabits-android/src/main/AndroidManifest.xml
@@ -148,6 +148,7 @@
 
         <activity
             android:name=".notifications.SnoozeDelayPickerActivity"
+            android:taskAffinity=""
             android:excludeFromRecents="true"
             android:launchMode="singleInstance"
             android:theme="@android:style/Theme.Translucent.NoTitleBar" />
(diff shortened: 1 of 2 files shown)


iSoron:
Thank you for looking into this, @hiqua. I have been testing the patch, and it looks like just removing the following line fixes the problem:
```kotlin
dialog!!.setOnDismissListener { finish() }
```
Removing this line does introduce a subtle issue; if you dismiss the dialog using the back button, then an invisible window now remains covering the entire screen. We need to hit the back button once again to get rid of it.

Anyway, this does make the root problem clear. It looks like `onPause` is being called after the user presses `later`, but before the user unlocks the screen. Because we currently call `dialog!!.dismiss()` in `onPause`, this causes the activity to finish. If this is really the problem, then another solution which seems to work is to just remove our `onPause` completely. I am not aware of any side effects, although there could be some.

In your patch, you also set `android:taskAffinity`. In what situation is this necessary?

hiqua:
Ok great, I've added a commit doing that then, I'll let you squash the commits. Let me know if there's something else I should test.

iSoron:
Thanks! It would be great if we could automatically test this in `large-tests`, but it does look quite hard.

Issue #843, opened 2021-03-12
unable to see task icons in app


alexjo82:
## Pre-submission checklist
- [x] I am submitting a bug report, not a feature request.
- [x] I am running the latest version of Loop Habit Tracker.
- [x] I have have searched for similar issues, but did not find any matches.

## Description
recent eye injury impact my vision. i switch to dark mode to help use this app but the icons for task blend in with background, double tried light mode and similar. light mode though hurts my eyes right now too

## Steps to reproduce
1. turn on dark mode
2. look at icons
3. try light mode again
4. look at icons

## System information
 - Phone: [e.g. Google Pixel 4] nexus 5
 - Phone Operating System: [e.g. Android 10] android 6
 - App version: [e.g. 1.8.9] 1.8

iSoron:
Duplicate of #413

PR #848, opened 2021-03-13
Add setting for higher contrast task symbols

melocene:
This PR should resolve the following:

#413 
#649 
#719 
#777
#843 

A new setting was added for a user to **opt-in** to using medium contrast for the checkmarks, X, and ? symbols for tasks. Other than the additional setting no other visual changes are made to the default state of the app. This should help with the concern of major change noted in #777 since again it is entirely **opt-in** for users who want or need the setting, existing users would see no change. No other UI elements seemed affected by this change.

I did also share the example screenshots with independent parties who did not feel the updated contrast clashed with the existing UI. To me the updated color also seems to more closely match the labels for the days as well, though due to them being larger they are sharper.

As much as I could I tried to keep variable and preference names similar to those already defined. I also ensured the text did not overflow in the settings menu, and placed it in the `Interface` section which seemed appropriate. I am allowing edits from the maintainers on this PR to adjust this text as they see fit for better clarity, or to fit their preferred style/wording.

Two issues I ran into:

1. My platform does not support running the tests from the bash script. Can I please get assistance with ensuring these run on this PR?
2. I do not speak languages other than English, so was not able to add translated strings. That being said I didn't see translation for the ? setting in those files either, so not sure this will be a big deal.

Screenshot of new setting:

screenshot_2021-03-13_12-03-28

Original low contrast color:

screenshot_2021-03-13_12-00-39
screenshot_2021-03-13_12-01-18

With higher contrast setting enabled:

screenshot_2021-03-13_12-01-46
screenshot_2021-03-13_12-01-36

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/CheckmarkButtonView.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/CheckmarkButtonView.kt
@@ -117,6 +117,7 @@ class CheckmarkButtonView(
     private inner class Drawer {
         private val rect = RectF()
         private val lowContrastColor = sres.getColor(R.attr.lowContrastTextColor)
+        private val medContrastColor = sres.getColor(R.attr.mediumContrastTextColor)
 
         private val paint = TextPaint().apply {
             typeface = getFontAwesome()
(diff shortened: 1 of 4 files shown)


hiqua:
1. as long as the GitHub actions run as they do currently, I think it's all good!
2. the translation is handled via Crowdin so indeed you don't have to do any of this.

iSoron:
Fixed in ee462715de.

I am not merging this PR since I would like to avoid introducing new options in the settings. This PR also fails to address my comments in #777. In usual circumstances, I would try to work with the author of the PR on revising it until it is ready to be merged. I am not doing it in this case because of the author's overall hostility towards myself and the project.

PR #874, opened 2021-04-06
Fix app crash on habit deletion when sorted not manually.

engineering4good:
As mentioned in [descussion](https://github.com/iSoron/uhabits/discussions/42#discussioncomment-246040), deletion of habit when sorted not manually (by name, color, etc.) crashes app.
This was fixed by check that the type of order is BY_POSITION (manual) before reordering.
All tests passed.

### DIFF
--- a/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/sqlite/SQLiteHabitList.kt
+++ b/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/sqlite/SQLiteHabitList.kt
@@ -131,9 +131,6 @@ class SQLiteHabitList @Inject constructor(private val modelFactory: ModelFactory
     @Synchronized
     override fun remove(h: Habit) {
         loadRecords()
-        if (Order.BY_POSITION.equals(list.primaryOrder)) {
-            reorder(h, list.getByPosition(size() - 1))
-        }
         list.remove(h)
         val record = repository.find(
             h.id!!
@@ -142,6 +139,7 @@ class SQLiteHabitList @Inject constructor(private val modelFactory: ModelFactory
             h.originalEntries.clear()
             repository.remove(record)
         }
+        rebuildOrder()
         observable.notifyListeners()
     }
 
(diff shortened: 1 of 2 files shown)


iSoron:
Thank you for the pull request, @engineering4good! I think we should have some automated tests to prevent this bug from coming back in the future. These could be either unit tests (in [SQLiteHabitListTest.kt](https://github.com/iSoron/uhabits/blob/55f3bab13a1854d66745becbe28a15af366c2cab/uhabits-core/src/jvmTest/java/org/isoron/uhabits/core/models/sqlite/SQLiteHabitListTest.kt)) or UI tests (in [acceptance/HabitTest.kt](https://github.com/iSoron/uhabits/blob/55f3bab13a1854d66745becbe28a15af366c2cab/uhabits-android/src/androidTest/java/org/isoron/uhabits/acceptance/HabitsTest.kt)). Perhaps both. Would you like to work on this? If not, please let us know.

engineering4good:
Modified test (in SQLiteHabitListTest.kt).

iSoron:
Thank you once again for the PR, @engineering4good, and welcome to the project. Unfortunately, while this patch does fix the crash, it seem to introduces another bug. Note that we must always update `record.position`, even if the list is currently sorted automatically. If we don't do that and the user later switches back to manual sorting, then the ordering will be corrupted. More precisely, we need to pass `SQLiteHabitListTest.testRemove`, regardless of the `habitList` ordering. The test below fails on both `dev` and after this patch:
```kotlin
   @Test
    fun testRemove_orderByName() {
        habitList.primaryOrder = HabitList.Order.BY_NAME_DESC
        val h = habitList.getById(2)
        habitList.remove(h!!)
        assertThat(habitList.indexOf(h), equalTo(-1))

        var rec = repository.find(2L)
        assertNull(rec)
        rec = repository.find(3L)!!
        assertThat(rec.position, equalTo(1))
    }
```

I think the best solution is to remove the `reorder` call entirely in `SQLIiteHabitList.remove` and instead call `rebuildOrder()` after the record has been deleted. This is a bit less efficient, but more safe.

Also, please revert the changes `testRemove` and add a new test instead.

engineering4good:
I'm using this beautiful app and happy to improve it. 😊

hiqua:
Naïve question: if `rebuildOrder` can replace `reorder` in this case, can we get rid of `reorder` altogether? No clue, but if they are somewhat equivalent, it would be better. The first has less code and looks code.

iSoron:
@engineering4good Thanks, merged!

@hiqua If I remember correctly, the `reorder` function is currently called repeatedly as the user drags a habit to a new position. For example, moving a habit up 5 positions on the list calls this method 5 times. If we call `rebuildOrder` every time, I would expect dragging to become much slower. If we could delay applying the changes to the database until after the user has moved the habit to its final position (that is, in the previous example, if we could reduce from 5 calls to 1 call), then your suggestion could work.

PR #894, opened 2021-05-02
Add missing dependency to github workflow

hiqua:
The logs after the latest changes were mentioning "flock command missing", this change installs it to prevent this.

I haven't looked into why flock is now required to prevent concurrency issues, but if it is we should make sure it's present.


iSoron:
Thanks! `flock` is required to prevent multiple processes from simultaneously calling `sdkmanager/avdmanager`.

Issue #194, opened 2016-10-21
Possibility to add "comment" or "Note" on a specific day ?


akhater:
I will try to give an example, I have as habit to sleep at least 7 hours / night and last night I had an overnight at work so, obviously, I failed this habit, but I would like to have a way to add a note to that day explaining why I failed it would help when looking back to understand what happened on a particular day

thanks

YPetremann:
Would be also be great for things like bloging or diversification habits

foobarbazquxfred:
It would also be nice if any numeric "note" left could be graphed. So a user could see how many minutes he meditated this month, for example.

I would be willing to contribute to a bounty for the note feature, FWIW.

Rightmanj:
It will also be useful for swimmers or runners, in order to keep truck the time and distance.

iSoron:
@Rightmanj This will be covered by numerical habits (see #42)

holocircuit:
I'd also like this, and would be willing to write it.

I'm not sure about the best way to do it with the interface though. Perhaps "super-long press" should bring up a menu with an option to add a note?

tickTackHack:
I'm also willing to give it a shot to implement this.  I would recommend making the input length restricted to a relatively short length. Maybe 240 characters? This can help with keeping the database size small and help deter security issues.

c33s:
not sure it makes sense to restrict the length of the field. i personally see no benefit on having a "small"  database. it looks like the performance impact is not that huge (should be tested on an android device before of course). the devices all have quite a lot of space, so keeping the size small does not really matter.  
even if someone writes a long text to 10 habbits, it is still only text. so it is not really big in size.
 
> I recognized, that performance drastically increases with table size dropping. I found:
> 
>     1.6M entries 327 ms w/o indexing => 29.7 ms with indexing
>     670k entries 149 ms w/o indexing => 28.8 ms with indexing
>     280k entries 71 ms w/o indexing => 28.5 ms with indexing
>     147k entries 44 ms w/o indexing => 28.0 ms with indexing
>     19k entries 25 ms w/o indexing => 25.0 ms with indexing
https://stackoverflow.com/q/33456278

> I've created SQLite databases up to 3.5GB in size with no noticeable performance issues. If I remember correctly, I think SQLite2 might have had some lower limits, but I don't think SQLite3 has any such issues.
> 
[...]

PR #1014, opened 2021-07-09
"Add note on a specific day" feature

aulia-adil:
Hello everyone, I want to solve this issue #194 and I have made the mockup. I use Figma and you can try it by clicking [here](https://www.figma.com/proto/GuIxtK7gGpn8nTERW7ESLu/loop-habit-tracker-reason-feature?node-id=12%3A1116&scaling=min-zoom&page-id=0%3A1) to see the prototype and [here](https://www.figma.com/file/GuIxtK7gGpn8nTERW7ESLu/loop-habit-tracker-reason-feature?node-id=0%3A1) to see the canvas.

Here is the explanation for my note feature. There are two parts: on the landing page and on the calendar.

Landing page flow

1. Double-tap the cross or check mark to display the input form
2. Write the note
3. Save

Calendar flow

1. Click edit
2. Press-and-hold the date to display the input form
3. Write the note
4. Save

The feedback for this feature is a small dot on the date on calendar.

@iSoron :

> Regarding the UI, how do you propose we display the notes, after entering them?

My solution is to create an input form that expandable to a certain size and the input form can be read and written at the same time. To see the input form, users have to open the calendar section and click on the date that has a note in it. Here is the illustration.

alt text
alt text
alt text

So here is my idea and please tell me what you think :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,3 +129,4 @@ contribute, even if you are not a software developer.
 [fdroid]: http://f-droid.org/app/org.isoron.uhabits
 [build]: https://github.com/iSoron/uhabits/blob/dev/docs/BUILD.md
 [beta]: https://play.google.com/apps/testing/org.isoron.uhabits
+ 


traxium:
We would certainly need "the feedback dots" on the main screen as well. Or we could use a little triangle in the upper right corner instead, as it's done in HabitHub/Google Sheets.

aulia-adil:
@traxium Like this?

traxium illustration

iSoron:
Thank you for the mockups, @aulia-adil. A few thoughts:

1. The idea of showing a dialog where the user can enter notes is fine. This is similar to what we currently have for numeric habits, and there has some discussion about creating a dialog/popup that allows the user to quickly switch between different checkmark values, instead of cycling through them (like [this demo](https://github.com/piruin/quickaction/blob/master/asset/demo.gif) than @chennemann shared in the chat). What I'm wondering is if we should spend some time to design a dialog that covers all three use cases. This is not needed for this PR, just something for us to keep in mind.
2. While having a dialog is fine, I am much less sure about how to show it. Double tap doesn't really work because it doesn't play well with the "toggle with short press" option.
   
   Something that could possibly work is: (1) if "toggle with short press" is enabled, then a single tap toggles, while a long press shows the dialog; (2) if "toggle with short press" is disabled, we flip these two: a single tap shows the dialog, while a long press toggles. I believe this is what @chennemann suggested before.
   
   An issue with this is that: (1) for users with "toggle with short press" enabled, the feature is not discoverable; (2) for users with "toggle with short press disabled", the quick toggle feature is not discoverable.
3. I was going to ask you about the note indicator for the main screen, but I see that @traxium already mentioned it above. Your solution looks good. I think we are only missing the indicator in the calendar chart now. Could you please create a quick mockup for that? A dot may work, but it may be too small. Another idea is adding a small triangle to the corner, [similar to Excel](https://www.gemboxsoftware.com/spreadsheet/examples/208/content/Comments.png). Note that "skips" are represented in the calendar as a hatch pattern, which may make the notes marker hard to see. Unfortunately, I imagine users would like to add notes precisely to days they skipped, so we need a good solution. Also note that the background color of the square changes according to status.

In any case, I think the mockups look good and the feature would be interesting to have. Please feel free to move ahead with the implementation, and please let us know if you have any questions. Here is a suggested list of subtasks. For now, let's not worry about numeric habits.

- [ ] Add a `notes` column to the `Repetitions` table. You will need to [create a new migration](https://github.com/iSoron/uhabits/tree/43921721d403f6f90b59f60dcf411a2d846d936f/uhabits-core/src/jvmMain/resources/migrations).
- [ ] Add a `notes` field to `EntryRecord` and `Entry`
- [ ] Implement the dialog for boolean habits
- [ ] Show the dialog in `ListHabitsActivity`
- [ ] Show the dialog in `HistoryEditorDialog`
- [ ] Make `CheckmarkButtonView` and `NumberButtonView` display the notes indicator
- [ ] Make  `HistoryChart` display the notes indicator

aulia-adil:
@iSoron :

> if "toggle with short press" is disabled, we flip these two: a single tap shows the dialog, while a long press toggles.

I used to think that why some users use long press toggles because they hate to accidentally check/uncheck the day. If the note feature exist and to be shown with single tap, this will reduce their experience.

Beside, the signifier "press-and-hold to check or uncheck" must be gone because the single tap is used for the dialog. This might confuse the first time user about how to check/uncheck a day.

Because of this, I propose that if "toggle with short press" is disabled, then we use double tap to show the dialog. As for the "if toggle with short press enabled", yes I agree with you by using a long press.

But I think it might annoy users if the signifier is to pop-up twice when they double tap. Because of this, I don't think the signifier should pop-up when it has displayed.

For the "toggle with short press enabled" mode, then the instruction to open the note should be shown in the setting when that mode is about to be enabled. Hopefully, the users will read that instruction.

illustration

> A dot may work, but it may be too small. Another idea is adding a small triangle to the corner, similar to Excel.

Like this? Frankly speaking, it doesn't suite my taste... And I think we can just enlarge the dots to become as big as the triangle Excel. 

illustration solution

> Note that "skips" are represented in the calendar as a hatch pattern, which may make the notes marker hard to see.

I see that I forgot to overlook into this problem. Now I propose this solution to make the notes marker.

illustration

By the way, I have updated the [prototype](https://www.figma.com/proto/GuIxtK7gGpn8nTERW7ESLu/loop-habit-tracker-reason-feature?node-id=3%3A241&scaling=min-zoom&page-id=0%3A1&starting-point-node-id=12%3A1116&show-proto-sidebar=1) to make the notes marker in hatch pattern and to enlarge the dots. You can take a look at it if you want.

That's all what I have in mind. We can discuss more of this but I will follow your decision @iSoron because you are the owner.

iSoron:
Hi @aulia-adil, I spent some time preparing high-fidelity mockups for this feature. Please see below. I think dots, like you suggested, work well both on the main screen and in the calendar. Feel free to suggest any changes. I will comment about the other points in your post soon.

- [Figma Design File](https://github.com/iSoron/uhabits/files/6835046/Loop.Habit.Tracker_.Daily.Notes.zip)


## Main Screen

Light
Dark

## Detail Screen

Light
Dark

## Calendar Screen

Light
Dark

iSoron:
Hi @aulia-adil, any updates on this? Please let me know if you have any questions about the implementation.

aulia-adil:
Hello @iSoron thank you for following up.

Actually, this is my first android project and I'm still learning android UI. On the other hand, I'm also having a hard time working on my thesis and this is my priority right.

So far, there is no significant progress... However, I'm planning to take mobile development course next semester and I hope I could develop this PR while taking the course. Anyway, is there a deadline for this PR?

hiqua:
No, no deadline, no worries. But maybe he'll want to close it in the meantime.

iSoron:
I have summarized this discussion in #1074. New PRs implementing #1074 are very welcome.

PR #1015, opened 2021-07-10
Reduce checkmark widget minimum size for fine-grained home screen grids

duxovni:
On my device (a Pixel 3a configured with a 5x7 homescreen grid, more fine-grained than the default), the 40dp x 40dp minimum dimensions of the checkmark widget end up as 1x2 grid cells, with a *lot* of wasted vertical space. Reducing it to 32dp x 32dp lets me have a 1x1 checkmark widget, which looks fine and works fine while leaving me a lot more screen space to work with.

### DIFF
--- a/uhabits-android/src/main/res/xml/widget_checkmark_info.xml
+++ b/uhabits-android/src/main/res/xml/widget_checkmark_info.xml
@@ -19,8 +19,8 @@
   -->
 
 <appwidget-provider xmlns:android="http://schemas.android.com/apk/res/android"
-                    android:minHeight="40dp"
-                    android:minWidth="40dp"
+                    android:minHeight="32dp"
+                    android:minWidth="32dp"
                     android:initialLayout="@layout/widget_wrapper"
                     android:previewImage="@drawable/widget_preview_checkmark"
                     android:resizeMode="vertical|horizontal"


hiqua:
Can you add screenshots before / after?

duxovni:
A screenshot of my homescreen, with two uhabits checkmark widgets, each taking up 1x2 grid cells
^ With uhabits 2.0.2. It won't let me resize the widgets any smaller than 1x2.

A screenshot of my homescreen, with two uhabits checkmark widgets, each taking up 1x1 grid cells
^ With my patched version. I can now have 1x1 checkmark widgets.

iSoron:
Thank you for the pull request, @duxovni. Any reason for picking 32dp instead of, say, 28dp? I think we need to figure out at what point does the current widget become truly unreadable and set it to that.

duxovni:
32dp seemed to be just below the cutoff point where it became too big for one grid cell, and I was trying to make the most conservative change that fixed the issue since I don't really have other devices to test on. I agree that it would make the most sense to figure out the practical minimum value if possible, though. You're more familiar with the relevant layout constraints than I am, I'm certainly happy to accept suggestions for smaller values.

hiqua:
If 1x1 is too small, we still have the option to enlarge the widget to 2x2 or something, right? So I'd say let's go to the minimum reasonable and we can revisit later, we're not committing too much by changing this value now.

Unfortunately I don't seem to be able to adjust the grid rows and columns in my AVDs so I wasn't able to test it that much myself.

iSoron:
I did some tests, and here are what 40dp x 40xp and 32dp x 32dp widgets look like:

widget-40dp-32dp

1. We cannot safely reduce the minimum height and width to 32dp because this is not enough space to render the widget.
2. 40dp x 40dp is the recommended size for 1x1 widgets, according to the official [Android documentation](https://developer.android.com/guide/practices/ui_guidelines/widget_design.html#anatomy_determining_size).

It looks like 32dp currently works for Pixel Launcher, but there is no guarantee that this would work for other launchers.

duxovni:
The current behavior is broken on my system, though. Does setting the minimum size as 32x32dp prevent users from enlarging it to whatever size is usable on their own systems?

duxovni:
I've updated my PR to leave minHeight and minWidth as-is, but add minResizeHeight and minResizeWidth set to 32dp. That way, the default widget size is unchanged from the current behavior, but it's still possible to resize it smaller if that's what works with your launcher. Is that acceptable?

duxovni:
Or rather I've pushed that change to my PR branch, but it doesn't look like Github will let me update a closed PR? In any case, [you can see the new proposed change here](https://github.com/duxovni/uhabits/commit/e52c1ed3f33d50df8cde0320d39183194f0d6594).

iSoron:
> The current behavior is broken on my system, though

@duxovni Please feel free to open an issue. I am closing this PR as simply reducing min height/width is not a proper solution. It works in your system, but it may break in many others. If we want to reduce min height/width to 32dp, then we need to ensure that the widget actually renders properly in these dimensions.

PR #1030, opened 2021-07-30
Use notification timestamp when pressing "no"

### LINKED ISSUE #969: Popup marks the wrong day
## Pre-submission checklist

## Description
When a popup is answered the following morning, it will mark the new day instead of the one the reminder refers to

## Steps to reproduce
0. Have a habit "brush teeth in the evening" with the question "Have you brushed your teeth this evening?" and a reminder at 2200
1. Observe the next morning that there is a reminder "Have you brushed your teeth this evening? Yes, no, Later"
2. Click on No because you know the reminder refers to last night, and you haven't.
3. Look at resulting habit list
4. Observe that yesterday remains at ? (no answer) but today was marked X for fail even though it wasn't due yet (note, long click on the wrongly entered X can only toogle between check and fail, it cannot revert to not-yet-decided which I would like to set it to because I don't know yet whether I will brush my teeth today in the evening - that's not the subject of this bug report but if I could have done so, I would have regarded it as cosmetic because a workaround exists and not bothered reporting it).

## System information
 - Phone: OnePlus 3
 - Phone Operating System: LineageOS (Android 9)
 - App version: 2.0.2

## Screenshots
not necessary

hiqua:
Fixes #969.

The logic was different from the one when pressing "yes", I guess from the days when pressing was a no-op anyway (apart from removing the notif).

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/intents/PendingIntentFactory.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/intents/PendingIntentFactory.kt
@@ -63,13 +63,14 @@ class PendingIntentFactory
             FLAG_UPDATE_CURRENT
         )
 
-    fun removeRepetition(habit: Habit): PendingIntent =
+    fun removeRepetition(habit: Habit, timestamp: Timestamp?): PendingIntent =
         getBroadcast(
             context,
             3,
             Intent(context, WidgetReceiver::class.java).apply {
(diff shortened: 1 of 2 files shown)


iSoron:
Thanks, @hiqua! I have applied it to branch `hotfix/2.0.3`. See 271de59a94ae540149fd60c293dffdcdcfb98f08

PR #1052, opened 2021-08-04
Hide completed and failed

hiqua:
Not sure how much you'll like the first commit. I can split this PR in two if that requires more discussion. The behavior change itself is relatively small (see second commit).

See #1049.

### DIFF
--- a/uhabits-android/src/androidTest/java/org/isoron/uhabits/HabitFixtures.kt
+++ b/uhabits-android/src/androidTest/java/org/isoron/uhabits/HabitFixtures.kt
@@ -23,10 +23,10 @@ import org.isoron.uhabits.core.models.Entry.Companion.YES_MANUAL
 import org.isoron.uhabits.core.models.Frequency
 import org.isoron.uhabits.core.models.Frequency.Companion.DAILY
 import org.isoron.uhabits.core.models.Habit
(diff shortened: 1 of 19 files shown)


iSoron (changes requested):
Thank you for the PR, @hiqua.

1. The refactoring is fine, but could you please submit it as a separate PR?
2. Could you please add some tests for `isFailedToday`?
3. If question marks are enabled, we also need to modify the menu message from "Hide completed" to "Hide entered" or something similar.

hiqua:
1. done
2. should I have different tests for isFailed or should I just insert assertions in the isCompleted tests that are already in the codebase? I was copy-pasting the tests until I realized you might prefer that I just add assertions inside the existing tests.
3. ok trying to do that

iSoron:
> should I have different tests for isFailed or should I just insert assertions in the isCompleted tests that are already in the codebase?

Either way is fine.

>  just because the unknown option is disabled, does not mean that the last tick is not a "no", since it could have been ticked while the option was active and before the user deactivated it again.

I don't think this is a big issue. Most users will probably not keep switching back and forth between "hide/show question marks". I'd like to keep "hide completed" for users that have question marks disabled because I think it's much more clear.

iSoron:
Looks good now, thanks!

PR #1054, opened 2021-08-05
Introduce HabitType and NumericalHabitType enums

hiqua:


uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/HabitType.kt
```diff
@@ -0,0 +1,17 @@
+package org.isoron.uhabits.core.models
+
+import java.lang.IllegalStateException
+
+enum class HabitType(val value: Int) {
+    YesNo(0), Numerical(1);
```
iSoron, line 6: For consistency with existing enums, could you please make these upper case (`YES_NO`, `NUMERICAL`)?

iSoron (changes requested):
Thanks, @hiqua. I left just a minor comment below.

PR #1056, opened 2021-08-08
Use cross to represent auto tick

hiqua:


iSoron:
@hiqua Could you please explain the reasoning behind this change? It's unclear to me.

hiqua:
Sorry I went a bit fast, I should have explained more. It's simply to distinguish between YES_AUTO and NO.

Right now, there's no difference, both show an empty colored tick. With this change, NO will be an empty colored cross instead. 

The color is meant to convey the meaning that the strike is still fine, and the cross that the habit could still be ticked today. There might be better ways to represent this state, I just think we have to be able to distinguish between the two situations.

iSoron:
> It's simply to distinguish between YES_AUTO and NO. Right now, there's no difference, both show an empty colored tick.

We already distinguish between all checkmark states. In the screenshot below, for example, we have: `UNKNOWN`, `UNKNOWN`, `NO`, `SKIP`, `UNKNOWN`, `YES_AUTO`, `YES_AUTO`, `YES_MANUAL`, `UNKNOWN`, `UNKNOWN`.

Screenshot_1628445100

hiqua:
I actually thought that an entry for which the user explicitly entered "no" kept this "no", but it seems that it's automatically converted to "yes_auto" instead. I'll follow up in a discussion.

hiqua:
Follow-up there: https://github.com/iSoron/uhabits/discussions/96#discussioncomment-1146574

PR #1063, opened 2021-08-16
Updating links in README

ganesshkumar:
Updating http links to https links. By using http links from browser - all of them are getting redirected to https automatically except http://www.gnu.org/licenses, which displays a warning message that the connection is not secure. Switching to https makes the connection secure as well.


iSoron:
Thanks, @ganesshkumar!

Issue #922, opened 2021-05-22
Make numpad work better in number picker dialog 


iSoron:
## Description

(originally proposed by @hiqua in #900)

Our current number picker dialog (shown below) uses two NumberPickers to allow the user to either type a value or scroll, whichever is the most convenient. Unfortunately our current implementation has two shortcomings:

1. Only the first NumberPicker is accessible through the numpad. For example, typing "3.15" in the numpad actually enters the value "315" because the widget does not recognize the decimal separator. It is currently completely impossible to change the focus to the second `NumberPicker` and type a number.
2. It is impossible to type fractional numbers that are not multiple of 0.05 (for example, 3.14).

This task is to:
- [ ] Make both NumberPickers focusable.
- [ ] Make the second NumberPicker have 0.01 increments instead of 0.05.
- [ ] Ensure that typing "." in the numpad switches the focus from the first NumberPicker to the second and selects the current number, so that typing "3.14" works as expected.

Note that, in some languages, the decimal separator is "," or "٫" instead of dot. An ideal implementation would work with these as well.

## Screenshot

![Screenshot_1621644720](https://user-images.githubusercontent.com/4699657/119209786-56c74180-ba6e-11eb-89aa-280b07dd7b72.png)

sumanabhi:
Hey @iSoron,

I think I got some idea about this as well and looking to take this in my bucket of PRs.

hiqua:
We need to allow a bunch of separators on top of the locale default to switch from one number picker to the other, because keyboards might not respect this locale, see https://stackoverflow.com/questions/3821539/decimal-separator-comma-with-numberdecimal-inputtype-in-edittext#comment80982671_34256139

PR #1082, opened 2021-08-22
Make numpad work better in number picker dialog

hiqua:
#922

- [x] Make both NumberPickers focusable.
- [X] Make the second NumberPicker have 0.01 increments instead of 0.05.
- [x] Ensure that typing "." in the numpad switches the focus from the first NumberPicker to the second and selects the current number, so that typing "3.14" works as expected.

Currently, entering the decimal part via the numpad has no effect.

I'm trying to listen to key events to switch from one NumberPicker to the other, but I don't see any event for the non-digit (soft) keys. This might be relevant: https://developer.android.com/training/keyboard-input/commands (we cannot rely on such key events).

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/NumberPickerFactory.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/NumberPickerFactory.kt
@@ -33,6 +33,7 @@ import org.isoron.uhabits.R
 import org.isoron.uhabits.core.ui.screens.habits.list.ListHabitsBehavior
 import org.isoron.uhabits.inject.ActivityContext
 import org.isoron.uhabits.utils.InterfaceUtils
+import java.text.DecimalFormatSymbols
 import javax.inject.Inject
 import kotlin.math.roundToLong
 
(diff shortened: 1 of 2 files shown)


uhabits-android/src/main/res/values/strings.xml
```diff
@@ -228,4 +228,5 @@
     <string name="activity_not_found">No app was found to support this action</string>
     <string name="pref_midnight_delay_title">Extend day a few hours past midnight</string>
     <string name="pref_midnight_delay_description">Wait until 3:00 AM to show a new day. Useful if you typically go to sleep after midnight. Requires app restart.</string>
+    <string name="decimal_separator">.</string>
```
iSoron, line 231: I think we can use [DecimalFormatSymbols.getDecimalSeparator](https://docs.oracle.com/javase/7/docs/api/java/text/DecimalFormatSymbols.html#getDecimalSeparator()) for this.

hiqua:
I wanted to use a TextWatcher to check that a separator is entered, but the filters remove it before it reaches the watcher, hence the current solution. Curious what you'll think about it / whether you think there's a better way.

I don't think I've seen https://code.google.com/p/android/issues/detail?id=35482 at least not with this PR, when does it occur? In the meantime I've deleted the corresponding workaround, as it allowed dots in the fractional part. If we need to put it back, I need to find out how to prevent that.

Still not perfect / to follow up:
* for the fractional part, if only one digit is entered, it should be multiplied by 0.1 instead of 0.01
* if possible, we should probably force the keyboard to always be open, instead of closing it if the user starts scrolling

iSoron:
Thanks for the PR, @hiqua. I've tested it and it seems to be working quite well. 

> I don't think I've seen https://code.google.com/p/android/issues/detail?id=35482 at least not with this PR, when does it occur? In the meantime I've deleted the corresponding workaround, as it allowed dots in the fractional part. If we need to put it back, I need to find out how to prevent that.

I have also tried to reproduce the bug in your version of the code, but I couldn't. I think it's fine to remove the workaround.

PR #1096, opened 2021-09-05
Dialog's positive button has bigger font size now

### LINKED ISSUE #1093: tiny "ok" button in enter value window
## Pre-submission checklist

## Description
The "ok" button is small in the window to enter the number from notification.

## Steps to reproduce
1. Setup a habit with measurable target and with notification.
2. Wait for the notification to show up.
3. Click "enter" 
4. See a small "ok" button.

## System information
 - Phone: Oneplus 8
 - Phone Operating System: lineageos 18.1
 - App version: 2.0.3

## Screenshots

Screenshot_20210901-203558_Habits

kalina559:
Setting the 'OK' button size, instead of it being some default value. Fixes  #1093

### DIFF
--- a/uhabits-android/src/main/res/values/styles.xml
+++ b/uhabits-android/src/main/res/values/styles.xml
@@ -169,6 +169,7 @@
     </style>
 
     <style name="DialogButtonStyle" parent="Widget.MaterialComponents.Button.TextButton.Dialog">
+        <item name ="android:textSize">@dimen/regularTextSize</item>
         <item name="android:textColor">@color/grey_100</item>
     </style>
 
--- a/uhabits-android/src/main/res/values/styles.xml
+++ b/uhabits-android/src/main/res/values/styles.xml
@@ -169,7 +169,7 @@
     </style>
 
     <style name="DialogButtonStyle" parent="Widget.MaterialComponents.Button.TextButton.Dialog">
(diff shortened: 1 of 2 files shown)


hiqua:
Hey thanks for the contribution!

I'm not sure that's the right fix, since this doesn't change if you change the font size system-wide, while e.g. "change value" does change. Could you try to make it so that it just scales normally like the other text?

hiqua:
Maybe even more important before you go further with that: can you reproduce the original issue? I can't, I'll add a comment on the issue.

kalina559:
You're totally right, now it scales along with other texts. I was able to reproduce it on Android 11

uhabits-android/src/main/res/values/styles.xml
```diff
@@ -169,6 +169,7 @@
     </style>
 
     <style name="DialogButtonStyle" parent="Widget.MaterialComponents.Button.TextButton.Dialog">
+        <item name ="android:textSize">@dimen/regularTextSize</item>
```
hiqua, line 172: You have an extra space after `name`.

hiqua (changes requested):

iSoron:
Thank you for the PR and welcome to the project, @kalina559. The fix looks good now.

PR #1100, opened 2021-09-08
Replaced .png drawables with vector assets where possible

### LINKED ISSUE #701: Replace bitmap icons by vector drawables
## Description

Loop currently uses bitmap images for all icons (for example, `ic_action_add_dark.png`). We should switch to vector drawables, which occupy less disk space. This would also allow us to have a single file for both light and dark themes.

Originally suggested in #673

## Implementation

Unless I am mistaken, all of our icons come from the clip art gallery from Android Studio, so vector versions of them can be generated using `File / New / Vector Asset / Clip Art`. If some icons are not there, feel free to replace them. We are already using [FontAwesome icons](https://fontawesome.com/) in other parts of the app, so those could be used here too.

The slightly more tricky part is making the same icon work for both light and dark themes. This can probably be achieved using `android:tint`.

## References

* https://developer.android.com/guide/topics/graphics/vector-drawable-resources
* https://medium.com/android-dev-hacks/android-vector-drawables-bfb515ba8f2e
* https://medium.com/androiddevelopers/using-vector-assets-in-android-apps-4318fd662eb9

kalina559:
All the icons that could be found in Android Studio clipart gallery have been replaced with their .svg versions. Resolves #701


hiqua:
We have to check how that works license-wise, given that some of these files are under the Apache license with LHT is under GPL3.

kalina559:
The ic_notification.xml was actually created by converting the intro_icon_1.png to .svg using an online converter, so I'm not sure which license applies to it.

kalina559:
And I also found this: https://www.apache.org/licenses/GPL-compatibility.html

iSoron:
Thank you for the PR, @kalina559. It looks good. I'll merge after the checks are done.

On a side note, I have just realized that many of these icons are no longer used within the app. If you would like to submit a PR removing these unused resources, that would also be very welcome.

kalina559:
Should I do it in a separate PR or just add it to this one?

iSoron:
@kalina559 That can be a separate PR. I've just merged this one. Thanks again!

PR #1102, opened 2021-09-12
Added delay after toggling a habit and before sorting

### LINKED ISSUE #1055: Introduce a small delay between toggling a check mark and hiding the habit or sorting the list
## Description

When "hide completed" is enabled, the app currently hides the habit as soon as the user toggles a check mark, making it impossible to mark the habit as failed or skipped. We should introduce a small delay between toggling a check mark and hiding the habit, so that the user can toggle through different states.

Originally suggested in #1053.

## Implementation

One potential way to implement this is to modify `CheckmarkButtonView.performToggle`. Instead of calling `onToggle` immediately, we could wait a few seconds before calling it.

```kotlin
    // CheckmarkButtonView.kt
    fun performToggle() {
        value = if (preferences.isSkipEnabled) {
            Entry.nextToggleValueWithSkip(value)
        } else {
            Entry.nextToggleValueWithoutSkip(value)
        }
        onToggle(value) // Delay this call
        performHapticFeedback(HapticFeedbackConstants.LONG_PRESS)
        invalidate()
    }
```

kalina559:
I'm not sure about the exact delay lengths, these seems fine though. Also, should I add any delay after clicking the "Hide completed" checkbox? Resolves #1055

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsMenu.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsMenu.kt
@@ -20,6 +20,8 @@
 package org.isoron.uhabits.activities.habits.list
 
 import android.content.Context
+import android.os.Handler
+import android.os.Looper
 import android.view.Menu
 import android.view.MenuInflater
 import android.view.MenuItem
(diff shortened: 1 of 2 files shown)


uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsMenu.kt
```diff
@@ -122,31 +124,40 @@ class ListHabitsMenu @Inject constructor(
             }
 
             R.id.actionSortColor -> {
-                behavior.onSortByColor()
+                { behavior.onSortByColor() }.delay(SORT_DELAY_MILLIS)
                 return true
             }
```
iSoron, line 129: We don't need to delay the menu actions; changing the order should be done immediately. Could you please remove all calls to `delay` in this file?

uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/CheckmarkButtonView.kt
```diff
@@ -80,15 +82,24 @@ class CheckmarkButtonView(
...
+            {
+                onToggle(value)
+                performHapticFeedback(HapticFeedbackConstants.LONG_PRESS)
+                invalidate()
+            }.delay(TOGGLE_DELAY_MILLIS)
+        } else {
+            onToggle(value)
+            performHapticFeedback(HapticFeedbackConstants.LONG_PRESS)
+            invalidate()
+        }
```
iSoron, line 102: 1. We can apply the delay in all cases (not only when skips are enabled).
2. I just realized that this solution unfortunately breaks the ripple effect. The ripple effect should play immediately after the checkmark is entered, but it's currently playing seconds after. I think a better place to add the delay might be around line 143 of HabitCardView.kt.

iSoron (changes requested):
Thanks for the PR, @kalina559.

iSoron:
Closing the PR due to inactivity. If you are still working on this, please let me know and I can reopen it.

Issue #1075, opened 2021-08-21
Remove Java dependencies from uhabits-core


iSoron:
## Description

In previous refactorings, we moved most of the application code from `uhabits-android` (a module with dependencies to Android classes) to `uhabits-core` (a Java module without any Android dependencies). While this made the application more portable and much faster to test, it is still not sufficient to make it run on iOS or the web. To leverage Kotlin Multiplatform, we need to remove all dependencies not only to Android, *but also to Java classes*! For example, any Kotlin classes that import `java.util` or `java.math` cannot be reused in other platforms.

This task is to remove Java dependencies from as many classes in `uhabits-core` as possible.

## Notes

1. As we remove Java dependencies, we should move the classes from `uhabits-core/src/jvmMain` to `uhabits-core/src/commonMain`. This will also prevent Java dependencies from being accidentally reintroduced in future code changes.
1. Many classes currently depend (directly or indirectly) on `java.util.Calendar` and `java.util.Date`. These classes should switch to `org.isoron.platform.time.LocalDate`, a pure-Kotlin implementation.
2. We already have multiplatform database support in `uhabits-core-legacy`. In includes an abstract [Database](https://github.com/iSoron/uhabits/blob/dev/uhabits-core-legacy/src/main/common/org/isoron/platform/io/Database.kt) class, as well as concrete [iOS](https://github.com/iSoron/uhabits/blob/dev/uhabits-core-legacy/src/main/ios/org/isoron/platform/io/IosDatabase.kt), [JavaScript](https://github.com/iSoron/uhabits/blob/dev/uhabits-core-legacy/src/main/js/org/isoron/platform/io/JsDatabase.kt) and [JVM](https://github.com/iSoron/uhabits/blob/dev/uhabits-core-legacy/src/main/jvm/org/isoron/platform/io/JavaDatabase.kt) implementations. This code needs to be ported to `uhabits-core` and the current `database/*.kt` files need to be removed.
[...]

PR #1108, opened 2021-09-19
Place MidnightTimer under test

sgallese:
MidnightTimer will need to move from executor to coroutine in order to complete Issue #1075
Prepares MidnightTimer to be testable with coroutine.
Requires removal of JVM dependencies from DateUtils before we can convert MidnightTimer though.

### DIFF
--- a/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/utils/MidnightTimer.kt
+++ b/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/utils/MidnightTimer.kt
@@ -39,7 +39,7 @@ open class MidnightTimer @Inject constructor() {
 
     @Synchronized fun onPause(): MutableList<Runnable>? = executor.shutdownNow()
 
-    @Synchronized fun onResume(delayOffsetInMillis: Long = DateUtils.MINUTE_LENGTH, testExecutor: ScheduledExecutorService? = null) {
+    @Synchronized fun onResume(delayOffsetInMillis: Long = DateUtils.SECOND_LENGTH, testExecutor: ScheduledExecutorService? = null) {
         executor = testExecutor ?: Executors.newSingleThreadScheduledExecutor()
         executor.scheduleAtFixedRate(
             { notifyListeners() },
--- a/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/utils/MidnightTimer.kt
+++ b/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/utils/MidnightTimer.kt
@@ -33,13 +33,19 @@ open class MidnightTimer @Inject constructor() {
     private val listeners: MutableList<MidnightListener> = LinkedList()
     private lateinit var executor: ScheduledExecutorService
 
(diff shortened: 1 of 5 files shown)


uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/utils/MidnightTimer.kt
```diff
@@ -39,11 +39,11 @@ open class MidnightTimer @Inject constructor() {
...
     @Synchronized fun onPause(): MutableList<Runnable>? = executor.shutdownNow()
 
-    @Synchronized fun onResume() {
-        executor = Executors.newSingleThreadScheduledExecutor()
+    @Synchronized fun onResume(delayOffsetInMillis: Long = DateUtils.MINUTE_LENGTH, testExecutor: ScheduledExecutorService? = null) {
+        executor = testExecutor ?: Executors.newSingleThreadScheduledExecutor()
         executor.scheduleAtFixedRate(
             { notifyListeners() },
-            DateUtils.millisecondsUntilTomorrowWithOffset() + 1000,
+            DateUtils.millisecondsUntilTomorrowWithOffset() + delayOffsetInMillis,
```
iSoron, line 46: Any reason for increasing the delay from 1000 ms to one minute?

uhabits-core/src/jvmTest/java/org/isoron/uhabits/core/utils/MidnightTimerTest.kt
```diff
@@ -0,0 +1,45 @@
...
+                        continuation.resume(true)
+                    }
+                }
+
+                MidnightTimer().apply {
+                    addListener(listener)
+                    // When
+                    onResume(1, executor)
+                }
+            }
```
iSoron, line 39: If you change the definition of `MidnightListener` to:
```kotlin
fun interface MidnightListener {
    fun atMidnight()
}
```
then this block of code can be simplified to:
```kotlin
            val suspendedListener = suspendCoroutine<Boolean> { continuation ->
                MidnightTimer().apply {
                    addListener { continuation.resume(true) }
                    // When
                    onResume(1, executor)
                }
            }
```

iSoron (commented):
Thank you for the PR, @sgallese. I left two minor comments below. This class is actually quite tricky to test, since we cannot wait 24 hours. Your test is a good start, though.

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/utils/MidnightTimer.kt
sgallese: Good catch, my mistake! 3e6a9181

uhabits-core/src/jvmTest/java/org/isoron/uhabits/core/utils/MidnightTimerTest.kt
sgallese: Smart a29943e

iSoron:
Looks good now, thanks!

PR #1113, opened 2021-09-22
Add border to create habit buttons

KristianTashkov:
Based on the discussion here https://github.com/iSoron/uhabits/discussions/1105, @MechtEngineer made a good point that on a pure black theme the buttons in create a new habit don't really look like buttons because their background and the actual background are both pure black.
I'm adding a border of the button with the text color to highlight the boundary more clearly. This is what it looks like on different theme options:
image
image
image


MechtEngineer:
Well done! That looks superb!

iSoron:
Looks good. Thanks, @KristianTashkov!

PR #1116, opened 2021-09-23
Implement skipPressed for Intro, update to latest version of AppIntro

sgallese:
For the app intro, pressing the Skip button would do nothing. Added support to finish the activity.
Also updated to the latest version of AppIntro using this migration doc: https://github.com/AppIntro/AppIntro/blob/master/docs/migrating-from-5.0.md

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -20,5 +20,6 @@ allprojects {
         maven(url = "https://oss.sonatype.org/content/repositories/snapshots/")
         maven(url = "https://kotlin.bintray.com/ktor")
         maven(url = "https://kotlin.bintray.com/kotlin-js-wrappers")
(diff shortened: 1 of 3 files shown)


sgallese:
Perhaps this update bumps a transitive dependency used by the HistoryWidgetTest.
But I'm not sure- it could simply be flaky and I'm not able to re-run the Github action.

https://github.com/iSoron/uhabits/blob/dev/docs/TEST.md#running-medium-tests
looks to have some old documentation not reflective of the latest build.sh file.

After some digging through the test code of build.sh and hacking things a bit, I was able to get `./build.sh android-tests 27` to run locally and all pass.

hiqua:
Yes we have some flaky tests unfortunately.

PR #1120, opened 2021-09-25
Remove JVM from DateUtils

sgallese:
**PR IN PROGRESS** - Showing migration work in case any obvious issues

For #1075

Starts to remove JVM dependencies from DateUtils, migrate to common LocalDate

Depends on https://github.com/iSoron/uhabits/pull/1111 (will rebase on dev once merged)

Next up

- [ ] Finish tests for Timestamp.kt
- [ ] Move Timestamp/Test.kt from jvm to core
- [ ] Remove fluid-locale


sgallese:
Odd gradle error on CI test failure.
I was able to run `:uhabits-core:allTests` task locally without issue.

sgallese:
I need to investigate test failure!

uhabits-core/build.gradle.kts
```diff
@@ -30,6 +30,8 @@ kotlin {
             dependencies {
                 implementation(kotlin("stdlib-common"))
                 implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core-common:1.3.8")
+                implementation("org.jetbrains.kotlinx:kotlinx-datetime:0.2.1")
+                implementation("io.fluidsonic.locale:fluid-locale:0.10.0")
```
sgallese, line 34: @iSoron OK to use this library for Locale? Or do we need to role our own per platform?

iSoron: It looks like the library's main functionality is parsing language tags (e.g. "ZH-HANT-cn-somevar"). Unless I am missing something, I don't think we need this functionality; we can just store the tag as is. For example, it seems to me like the following `Locale` class would suffice for our purposes, but please let me know if I'm mistaken:

```kotlin
data class Locale(val tag: String)
```

hiqua:
@sgallese are you still working on this?

sgallese:
Unfortunately this task seems to continually grow :( I'm thinking I should
abandon in favor of smaller, quicker wins on removing the JVM. Ok to close
this PR.

PR #1147, opened 2021-10-03
Added delay after toggling a habit and before sorting, update

hiqua:
Follow-up from #1102, with comments addressed.

@kalina559 FYI, feel free to take things from there if you wish to continue, otherwise I'll try to get this merged.


hiqua:
If you have an opinion on the best way to avoid sleeping in these particular tests I can look into it.

iSoron:
> If you have an opinion on the best way to avoid sleeping in these particular tests I can look into it.

Just a random thought, but maybe we could use a multiple of some animation duration, instead of 2000 ms? AFAIK, during test, the duration of all animations are set to zero, so the delay would disappear. In any case, I don't think this is a big issue, since large tests are already very slow.

iSoron (approved):

hiqua:
Ok, `val delay = 4 * context.resources.getInteger(android.R.integer.config_longAnimTime).toLong()` could work, but it's not set to 0 during the test. Maybe I'll revisit at some other time, but if it's not a concern I'll just merge how it is now.

PR #1177, opened 2021-11-05
Use the value of the Entry during CSV export

### LINKED ISSUE #1162: Export to CSV has additional commas that break the import to Excel or similar
Love that you can export data for your own analysis.

However, the export format is a bit... unrefined? Each entry restates the date, followed by a comma, followed by the actual value. This causes import to Excel and similar files to break, because every comma is considered a new column. 

I can write a regex to change them. But it's a csv export, right?

hiqua:
Fixes #1162.

Also test the content of the csv files after export.


iSoron:
Looks good, thanks!

PR #1178, opened 2021-11-05
Delete unused test resources

hiqua:
These resources seem to be duplicated, especially given that the tests pass without them, or am I missing something?


iSoron:
@hiqua I think you're right. Thanks!

Issue #1179, opened 2021-11-05
Measurable habits are always hidden if "hide completed" is ticked (on dev)


hiqua:
## Pre-submission checklist
- [X] I am submitting a bug report, not a feature request.
- [X] I have searched for similar issues, but did not find any matches.

## Description
See title.

## Steps to reproduce
1. use the latest version on `dev
2. create a new habit measurable habit with a target of 10
3. check "hide completed"
4. the habit seems to be hidden although it has no data yet

## System information
 - App version: `dev`

Setting `cant-reproduce` label until somebody else confirms I'm not missing anything.

iSoron:
@hiqua Thanks, I see the same issue here.

PR #1181, opened 2021-11-06
Fix "hide completed" for numerical habits

iSoron:
Bug #1179 was introduced in 7fe3ce970c42015305cb007b826d9e4d117cfdc5. It appears that the main problem was the lack of clarity in the meaning of `HabitMatcher.isCompletedAllowed`. This PR fixes the issue by introducing the following changes:

1. Replace `Habit.isFailedToday` by `Habit.isEnteredToday`
2. Add `isEnteredAllowed` in `HabitMatcher` and simplify the check
3. In `ListHabitsMenuBehavior`, set different filters according to `preferences.areQuestionMarksEnabled`

@hiqua As you wrote the original commit and identified the issue, would you like to review this?


hiqua:
Looks good, thanks!

hiqua:
Follow-up: would it make sense to split numerical habits and yes/no habits into different classes?

PR #1184, opened 2021-11-07
Don't unfocus after select 'every day' frequency

### LINKED ISSUE #1182: Cannot select "Every day" in frequency popup
## Pre-submission checklist

## Description
When attempting to change the frequency of a habit, the popup has an option for "Every day" but it cannot be selected

## Steps to reproduce
1. Create new habit
2. Click on Frequency
3. Choose "Every day"
4. Selection automatically goes back to "Every X days"

## System information
 - App version: [e.g. 2.0.3]

## Screenshots
habits-bug2

Workaround is to choose "Every [1] days", which is the same thing.  But it is confusing to have an option that cannot actually be selected.

hiqua:
Fixes #1182.


hiqua:
The second commit is simply because it seems to work without these calls to `requestFocus()`, so why do them?

iSoron:
Thanks, @hiqua!

PR #1186, opened 2021-11-08
Start the cumulative numerical habit implementation

hiqua:
This implementation doesn't work, because we then only show the sum of the past few days instead of today's value.

See the TODO in the code, but I think that we need another `value` field in the Entry.

I think at this point it would make sense to more cleanly separate between numerical and yes/no habits, e.g. have a proper `NumericalEntry` class. This will probably require a lot of changes in the data model though, so I was curious about your opinion about this.

### DIFF
--- a/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/EntryList.kt
+++ b/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/EntryList.kt
@@ -95,7 +95,27 @@ open class EntryList {
         clear()
         val original = originalEntries.getKnown()
         if (isNumerical) {
-            original.forEach { add(it) }
+            assert(frequency.numerator == 1)
+
+            val intervals = buildIntervals(frequency, original)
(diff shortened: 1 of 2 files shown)


hiqua:
I noticed that the "history" view in the details of the habit computes the cumulative score correctly, I should try to reuse this logic instead.

iSoron:
> See the TODO in the code, but I think that we need another value field in the Entry.

Perhaps we could add a boolean field `isCompleted`, which would eventually replace `habit.isCompletedToday()` and `value / 1000.0 >= targetValue` in the entire code base.

> I think at this point it would make sense to more cleanly separate between numerical and yes/no habits, e.g. have a proper NumericalEntry class. This will probably require a lot of changes in the data model though, so I was curious about your opinion about this.

If this refactoring makes the code more clear, it would be fine. Do you have a specific class/method that would particularly benefit from it?

iSoron:
Closing due to inactivity. If you are still working on this, please feel free to reopen.

PR #1278, opened 2022-01-20
Handle numerical habits from HabitBull

hiqua:


iSoron:
Thank you for the PR, @hiqua. I will review it soon.

iSoron:
Everything looks good, @hiqua. A bit surprising to see that the CSV doesn't have anything to indicate that the habit is numeric, but I think your idea here of checking `value > 1` is good. Thanks again!

PR #1319, opened 2022-03-09
Issue 1316: Skip measurable habit

### LINKED ISSUE #1316: Skip days not available for measurable habits
Pre-submission checklist
• [x] I am submitting a bug report, not a feature request.
• [x] I have searched for similar issues, but did not find any matches.

Description

I created a measurable habit.
I also have skip days enabled.
There is no way for skipping measurable habit.

Steps to reproduce
• long pressing a measurable habit opens up ticker. Which only allows you to select a number.
There is no way to select skip, even though it is enabled in settings

System information
• Phone: oneplus 
• Phone Operating System: Android 11
• App version: 2.0.3

Screenshots

kalina559:
Resolves #1316. My solution is to add a 'SKIP' button to the numerical habit dialog. If it's clicked, the entry value is set to Entry.SKIP / 1000 = 0.003 and it's displayed in the same way as a skipped yes/no habit.


uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/Entry.kt
```diff
@@ -27,7 +27,7 @@ data class Entry(
         /**
          * Value indicating that the habit is not applicable for this timestamp.
          */
-        const val SKIP = 3
+        const val SKIP = -2
```
hiqua, line 30: Why this change?

I think it's because of the other change where you want to match a whole range (`>0`), but I think it's not a good idea. This object is basically an enum, I think it's better to match every single case, so that it will be easier to see what needs to change if we add a new possible value.

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/ui/screens/habits/show/views/HistoryCard.kt
```diff
@@ -153,10 +153,15 @@ class HistoryCardPresenter(
...
                 entries.map {
                     when {
-                        it.value == Entry.UNKNOWN -> OFF
-                        (habit.targetType == AT_MOST) && (it.value / 1000.0 <= habit.targetValue) -> ON
-                        (habit.targetType == AT_LEAST) && (it.value / 1000.0 >= habit.targetValue) -> ON
-                        else -> GREY
+                        it.value == SKIP -> {
+                            HistoryChart.Square.HATCHED
+                        }
+                        it.value > 0 -> {
```
hiqua, line 159: As mentioned before, better to have an exhaustive match (every single value) rather than whole range (`>0`).

hiqua (changes requested):
Very nice, thanks!

hiqua:
The ktlint stuff can be a bit annoying, you can have a look at https://github.com/iSoron/uhabits/blob/dev/docs/GUIDELINES.md#code-style to make it less annoying. I use the commit pre-hook personally.

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/Entry.kt
kalina559: But if I leave it as it is, if user inputs 0.003 as an entry, it might mistake it for a skipped day

kalina559: Oh wait a second we use only 2 decimal places for the entries - you're right, I'll correct it later

iSoron: This change by itself (without a corresponding database migration) would also break all existing Loop databases. Since `0.003` cannot be entered by the user, I think we can use it for now. In the future, we should probably split this into two fields. For example, if the user enters a value, then `field1=YES_MANUAL`, `field2=500.0`. If the user skips, then `field1=SKIP`, `field2=<ignored>`.

iSoron:
Hi @kalina559, thank you for the PR. Besides the comments above, I have a few other questions:

1. For numerical habits with daily frequency, how does a `SKIP` differ from an entry with value 0? It appears to me that the difference, at this point, is purely visual. Shouldn't `SKIP` also have an impact on the score and the "target" chart?
2. For numerical habits with weekly or monthly frequency, does it even make sense to allow `SKIP`? When should the user enter this option, instead of entering zero? What's the impact on score and target in this case?

kalina559:
Hi @iSoron,
1. You're right, I'll look into how I can affect the score and target chart with this
2. User can enter SKIP in the history chart or in the reminder dialog. Maybe if all entries in a week/month are skipped, then the score and target are affected somehow?

iSoron:
> You're right, I'll look into how I can affect the score and target chart with this

For scores, I suggest copying the behavior from boolean habits: a `SKIP` keeps the score the same as the day before.

For the target chart, one thing that may make sense is to reduce the total. For example, if you have a daily numerical habit with target of 10, then the chart by default would show a monthly target of 300 (for a 30-day month). However, if you skip one day, the monthly target for that month could drop to 290. This might be difficult to implement, though. We would need to count the number of skips in a certain interval.

Regarding skips in non-daily numerical habits, I suggest disabling them for now until we understand that use case better. They don't seem to make a lot of sense to me at this point, but I'm open to suggestions.

ke352802081770314:
> Regarding skips in non-daily numerical habits, I suggest disabling them for now until we understand that use case better. They don't seem to make a lot of sense to me at this point, but I'm open to suggestions.

As you explained, skip should reduce the multi-day targets. so if I have a target "150 per week" and I have a skip day in that week, the target should be 6/7 =(7-skip_days)/7 of that number, so it is really what you already suggested, only the user-supplied value is on a different scale (weekly given directly instead of calculated as 7*daily etc.). 
I must admit scoring isn't that transparent to me so I can't explain how to adjust scoring in that case. My conceptual logic would be "skip days are treated like they didn't exist in the calendar" i.e. skipping 2022-03-13 would make the scoring think that the next day after 2022-03-12 is 2022-03-14 and everything else follows logically from that.

It could be argued that the skip supports laziness, because if I intended to achieve the cumulative number not by doing it daily anyway, skip could be used to lower the target. But honesty is always up to the user, and if I mark three days as skip because I was bedridden sick it is correct that my jogging target for that week should only be 4/7 of when I had been healthy.

uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/NumberButtonView.kt
```diff
@@ -181,6 +188,11 @@ class NumberButtonView(
                     typeface = BOLD_TYPEFACE
                     textSize = dim(R.dimen.smallTextSize)
                 }
+                value == Entry.SKIP.toDouble() / 1000 -> {
+                    label = resources.getString(R.string.fa_skipped)
+                    textSize = dim(R.dimen.smallTextSize)
+                    typeface = getFontAwesome()
+                }
```
iSoron, line 195: Skips are still rendering as zero to me. I think for this condition to apply, it must come before `when value >= 0`

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/ScoreList.kt
```diff
@@ -100,20 +100,25 @@ class ScoreList {
...
+                            )
+                        } else {
+                            if (normalizedRollingSum > 0) 0.0 else 1.0
+                        }
                     }
-                }
 
-                previousValue = compute(freq, previousValue, percentageCompleted)
+                    previousValue = compute(freq, previousValue, percentageCompleted)
+                }
```
iSoron, line 121: Could you please add some tests to ensure that scores are being computed correctly when skips are present?

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/ui/screens/habits/show/views/TargetCard.kt
```diff
@@ -51,37 +52,64 @@ class TargetCardPresenter {
                 isNumerical = habit.isNumerical
             ).firstOrNull()?.value ?: 0
 
+            val skippedDayToday = entries.countSkippedDays(
+                truncateField = DateUtils.TruncateField.DAY,
+                isNumerical = habit.isNumerical
+            ).firstOrNull()?.value ?: 0
+
```
iSoron, line 59: Not a required change, but this function is now looking very repetitive, especially after the changes introduced in this PR. Perhaps we could rewrite it.

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/EntryList.kt
```diff
@@ -301,6 +306,32 @@ fun List<Entry>.groupedSum(
     }.entries.map { (timestamp, entries) ->
         Entry(timestamp, entries.sumOf { it.value })
     }.sortedBy { (timestamp, _) ->
-        - timestamp.unixTime
+        -timestamp.unixTime
+    }
+}
+
+/**
+ * Counts the number of days with vaLue SKIP in the given perio
```
iSoron, line 314: Small typo at the end of the sentence.

iSoron (changes requested):
Thank you for the PR, @kalina559. I think it looks mostly good.

@hiqua Please free to add any comments.

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/ScoreList.kt
kalina559: Sure

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/ui/screens/habits/show/views/TargetCard.kt
kalina559: Maybe I could add a skippedDays: Int field to the Entry class, and then count these days in the List<Entry>.groupedSum?
My only issue with that is that this field would be only used when creating the TargetChart, so it wouldn't be great to have this almost never used field in the class 😄 
Another solution would be to create a class identical to Entry, but extended with the skippedDays field, create a .groupedSum method for it and use it in the TargetChart creation. It would be nice to just make a class that inherits from Entry, but I don't think it's best practice to use inheritance with data classes.
Should I try to implement any of these solutions?

iSoron: Let's just keep it like this for now.

iSoron:
Looks good to me. Thanks, @kalina559!

PR #1103, opened 2021-09-13
Add notes to specific dates

vbh:
closes #1074 

## Subtasks

- [x]     Add a notes column to the Repetitions table.
- [x]     Add a notes field to EntryRecord and Entry
- [x]     Implement the dialog for Yes/No habits
- [x]     Implement the dialog for Numeric habits
- [x]     Show the dialog in ListHabitsActivity
- [x]     Show the dialog in HistoryEditorDialog
- [x]     Make CheckmarkButtonView display the notes indicator
- [x]     Make NumberButtonView display the notes indicator
- [x]     Make HistoryChart display the notes indicator
- [x]     Make HistoryChart react differently to short and long press

## Images
<a href="https://imgbox.com/WwVa7rZq" target="_blank">image host</a> <a href="https://imgbox.com/6U3eieG3" target="_blank">image host</a> <a href="https://imgbox.com/soMnugja" target="_blank">image host</a> <a href="https://imgbox.com/yQo6w0Aa" target="_blank">image host</a>

PR #1346, opened 2022-04-12
Replace CheckmarkDialog by CheckmarkPopup

iSoron:
## Description

This PR replaces the `CheckmarkDialog` introduced in #1103 by a much smaller popup which shows up near the checkmark button, and that can be dismissed with a single click. See screenshots below.

@vbh @hiqua Please let me know if you have any comments or suggestions regarding the UI or the code. I think we could also have a version for numerical habits.

### After
Screenshot_1649725503

### Before
Screenshot_1649725569


uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/ui/screens/habits/list/ListHabitsBehavior.kt
```diff
@@ -65,15 +64,7 @@ open class ListHabitsBehavior @Inject constructor(
...
-            screen.showCheckmarkDialog(
-                entry.value,
-                entry.notes,
-                timestamp.toLocalDate(),
-                timestamp.toDialogDateString(),
-                habit.color,
-            ) { newValue, newNotes ->
-                commandRunner.run(CreateRepetitionCommand(habitList, habit, timestamp, newValue, newNotes))
-            }
+            throw RuntimeException()
```
hiqua, line 67: No better way here I guess, but that shows we should make a better distinction between numerical and yes/no habits (different classes).

hiqua:
* It might be slightly confusing that there's no "save" / "cancel" buttons for the text, if you don't want to change the mark, but just the text. I guess it's fine and users can get used to it, but it's also not like we need more space in that dialog. It also looks like you can't undo, e.g. if you delete your note by mistake. On the other hand, you don't want to discard the change if you tap outside of the window. This should also be consistent with the numerical habit dialog.
* the renamings / refactorings all merged into one commit make it harder to review unfortunately, maybe we should try to split that into several commits next time.

iSoron:
Thanks for the useful comments, @hiqua:

- I have split the PR into four commits. The first two include do some refactoring, the third one implements the main new functionality, and the last one removes the previous dialog. Some of the diffs in the last commit are confusing, but I am just deleting stuff.
- Your comment about the test made me realize that the new implementation was untestable, since we had random views creating dialogs from the middle of the view hierarchy. In the new implementation, the view just emits an `onEdit` event, which bubbles up. This is similar to how we handled the dialog before. We are still missing proper tests for the popup itself, though, but we never had them for the dialog either.
- I have made the notes box appear on top of the buttons. Is it now more clear you should first type your notes, then tap the checkmark state? I'm trying to keep the popup operable with two clicks for efficiency (one click to open, one click to select+save). If we add a separate save button, I'd say it's unexpected for the other buttons to close the popup; but then we would need three clicks.
- I'm also planning to create a version of this popup for the measurable habits, so consistency is not an issue right now.

Please let me know if you have any other comments.

hiqua:
> I have made the notes box appear on top of the buttons. Is it now more clear you should first type your notes, then tap the checkmark state?

I think it is, yes.

I'm wondering if the changing position is the right choice. Pros of a fixed centered position:
* predictible, i.e. easy to tap quickly into the note area / two options
* same typing method no matter the day / habit (think of a very large phones for which each thumb only reaches the left or right side)
* the code for it would likely be less fragile than the variable position, especially since there should be less code
* in particular, more likely that we'll have a consistent behavior across Android versions

Pros of the current variable position:
* easier to see for which habit / day the dialog has been opened

Up to you of course.

Other than that lgtm.

iSoron:
Thanks for the feedback, @hiqua. The reason for showing the picker close to the button is to make it behave like a popup, instead of a dialog, but you have a good point. Let's go with this for now, and if it turns out to be a problem in different phones or Android versions, we can revert to a fixed location.

PR #1350, opened 2022-04-14
Resort habit list after habit edit

### LINKED ISSUE #1131: Changing habit goals should trigger a sort refresh
## Pre-submission checklist

## Description
If we sort by status, the order depends on the goals, but changing habit goals does not trigger a sort refresh, meaning that the sort can become outdated after such a change.

## Steps to reproduce
1. sort by status (maybe by score as well? Not sure if that's impacted by a change of goals)
2. change a habit goal so that a habit that was not ticked becomes automatically ticked (e.g. because done yesterday and the habit is now once a week)
3. see that this habit hasn't changed its position

## System information
- App version: 2.0.3

hiqua:
Fixes #1131.

### DIFF
--- a/uhabits-android/src/androidTest/java/org/isoron/uhabits/acceptance/steps/CommonSteps.kt
+++ b/uhabits-android/src/androidTest/java/org/isoron/uhabits/acceptance/steps/CommonSteps.kt
@@ -178,6 +178,22 @@ object CommonSteps : BaseUserInterfaceTest() {
         EditHabitSteps.clickSave()
     }
 
+    fun changeFrequencyToDaily(habitName: String) {
+        clickText(habitName)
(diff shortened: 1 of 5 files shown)


iSoron:
Thanks, @hiqua. Would you like to write a regression UI test for this bug. We already have some in [ListHabitsRegressionTest](https://github.com/iSoron/uhabits/blob/2245347e28a218fd5b8cc3a618437276897f0e3e/uhabits-android/src/androidTest/java/org/isoron/uhabits/regression/ListHabitsRegressionTest.kt).

hiqua:
Done, added the commit before to make it easier to check it does fail without the check, please squash them before merging.

iSoron:
Thanks, @hiqua!

PR #1356, opened 2022-04-20
Issue 1332: Multiple numeric dialogs appearing

### LINKED ISSUE #1332: Multiple numeric dialogs appear after double clicking calendar tiles
## Pre-submission checklist

## Description
In the Calendar view of measurable habits, if user double clicks on a tile or quickly click on a couple of different tiles, multiple dialogs appear.

## Steps to reproduce
1. Go to measurable habit's calendar view
2. Click on a tile and then immediately click again on the same tile or any other tile
3. Dismiss the dialog
4. See that another dialog is below the first dialog.

## System information
 - Phone: Samsung A51
 - Phone Operating System: Android 11
 - App version: 2.0.3

kalina559:
Resolves #1332.
I'm not sure about this solution though, as it's a bit of a hack. The bug itself is not that serious - users can just enter the data twice, it's not breaking the app or anything. What do you think @iSoron @hiqua ?
While working on it I also noticed that the issue appears when clicking on the tiles in ListHabitsScreen, I corrected this there as well. Numeric dialog is also shown when using a Checkmark widget, but the issue wasn't appearing there.


iSoron:
Thanks for tackling this bug, @kalina559. I agree it's not serious, but it would still be nice to fix it.

I think keeping track of the number picker visibility in `NumberPickerFactory` is problematic. For example, if, for any reason, `onDismissListener` does not get called, we would get stuck in a state where the number picker can no longer be shown until the app is restarted.

I would suggest the following instead. Whoever calls `NumberPickerFactory(...).create` should keep a reference to the dialog, and should dismiss it before showing a new one. Something like this:

```
if (currentPicker != null) currentPicker.dismiss()
currentPicker = NumberPickerFactory(this@ShowHabitActivity).create(value, unit, notes, dateString, frequency, callback)
currentPicker.show()
```

kalina559:
I made something between the initial solution and the one you described - we're keeping a reference in a companion object of 
`NumberPickerFactory` and we dismiss it when create() is called to create a new dialog. This way we avoid writing almost identical code in both `ListHabitsScreen `and `ShowHabitActivity`.

The only downside to that is that users can see the first dialog for a split second before it gets dismissed. In the first solution the second dialog was prevented from being created if another one already existed, so the UX was a bit better.

uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/show/ShowHabitActivity.kt
```diff
@@ -173,7 +173,14 @@ class ShowHabitActivity : AppCompatActivity(), CommandRunner.Listener {
...
         ) {
-            NumberPickerFactory(this@ShowHabitActivity).create(value, unit, notes, dateString, frequency, callback).show()
+            NumberPickerFactory(this@ShowHabitActivity).create(
+                value,
+                unit,
+                notes,
+                dateString,
+                frequency,
+                callback
+            ).show()
```
iSoron, line 183: These changes look unnecessary. Could you please revert them?

iSoron (changes requested):
Thanks, that also works. I only have one comment below. Other than that, the PR looks fine to me.

kalina559:
Hey @iSoron, can I merge this?

iSoron:
@kalina559 I've merged it, thanks!

Issue #1104, opened 2021-09-17
Measurable doesn't mark complete


Lissidragon:
## Pre-submission checklist
- [ y] I am submitting a bug report, not a feature request.
- [y ] I have searched for similar issues, but did not find any matches. (There was one feature request but didn't seem exact match)

## Description
I created a measurable task to clean the office 60 mins every month. No matter how many times I put in 60 mins, the task isn't marked completed so it doesn't hide. 

## Steps to reproduce
1. Create a new measurable task.
2. Set units to mins. Set target to 60. Set frequency to month.
3. Go to Calendar or Main view to set several days to 60.
4. Return to main view if not there already.
5. Note that task does not hide upon completing 60mins within a months time.

## System information
 - Phone: Google Pixel 4
 - Phone Operating System: Android 11
 - App version: 2.0.3

## Screenshots
![Screenshot_20210917-110853](https://user-images.githubusercontent.com/11533909/133819983-022a6044-7fc0-4c6f-b148-1801eeea3d45.png)
![Screenshot_20210917-110923](https://user-images.githubusercontent.com/11533909/133819985-b7ca6abc-5611-40d1-b665-da0cc245bdcd.png)
![Screenshot_20210917-110931](https://user-images.githubusercontent.com/11533909/133819986-540141cc-137d-4d11-9c95-930c01c319a7.png)

sgallese:
@Lissidragon if you click the filter option, do you have "Hide completed" checked? When I have this option checked, the task is hidden for me once I complete the task.
![filter-option](https://user-images.githubusercontent.com/140911/133915719-375322c5-306c-4259-a705-87ec52e7f39a.png)
![hide-completed](https://user-images.githubusercontent.com/140911/133915720-3f68713f-a45e-41d9-8a3d-2371b1c791c0.png)

Lissidragon:
Yes, hide completed is checked. Monthly and weekly measurable won't go away
unless I do the required time that day. But it should go away if I
completed 60min that month overall or weekly should it not?

Elissa Pendragon

Lissidragon:
This "Test" is a monthly with 60min as target. See that hide completed is on and the task is still there despite 60min being done 3 days ago. (I toggled the hide but it still didn't go away)
![Screenshot_20210919-092321](https://user-images.githubusercontent.com/11533909/133931201-aaba6c53-5a8f-44bf-a40f-4760d70c05a0.png)

sgallese:
[...]

PR #1360, opened 2022-04-27
Measurable doesn't mark complete (#1104)

marcoebbinghaus:
Potential fix for confirmed bug https://github.com/iSoron/uhabits/issues/1104

for measurable habits: all entered values from the configured frequency are considered for completed-check (instead of only today's value).

Only supports frequencies DAILY, WEEKLY, MONTHLY (SelectItems from GUI for measurable habits)

### DIFF
--- a/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/Frequency.kt
+++ b/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/Frequency.kt
@@ -45,5 +45,8 @@ data class Frequency(
 
         @JvmField
         val WEEKLY = Frequency(1, 7)
+
+        @JvmField
+        val MONTHLY = Frequency(1, 30)
     }
 }
--- a/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/Habit.kt
+++ b/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/Habit.kt
@@ -73,7 +73,7 @@ data class Habit(
     }
 
     private fun fetchFirstDayToConsider(): Timestamp {
-        return when(frequency) {
+        return when (frequency) {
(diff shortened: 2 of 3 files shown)


hiqua:
@marcoebbinghaus Maybe you've seen it, but `./gradlew ktlintFormat` can take care of the formatting for you.

marcoebbinghaus:
> @marcoebbinghaus Maybe you've seen it, but `./gradlew ktlintFormat` can take care of the formatting for you.

Yup, found it. Thank you 👍

hiqua:
I haven't spent much time, but it does look reasonable! It'd be nice to also have a test for this (that would fail before this PR and succeed after). You could try to follow the examples in https://github.com/iSoron/uhabits/blob/dev/uhabits-android/src/androidTest/java/org/isoron/uhabits/regression/ListHabitsRegressionTest.kt

I'm not sure if there are helpers to deal with numerical habits though.

iSoron:
Thank you for the PR, @marcoebbinghaus. It looks well written and it solves the problem partially, but unfortunately this approach doesn't allow us, for example, to highlight in the calendar what days have already been completed due to previous repetitions, like we currently do for yes/no habits. I think a proper solution would be, for each `Entry`, to record not only the current value, but also the cumulative value. I am also concerned about the performance of this solution; it might not be a problem here, but in general we should not recalculate values every time we query them.

iSoron:
Thank you for the PR, @marcoebbinghaus. It looks well written and it solves the problem partially, but unfortunately this approach doesn't allow us, for example, to highlight in the calendar what days have already been completed due to previous repetitions, like we currently do for yes/no habits. I think a proper solution would be, for each `Entry`, to record not only the current value, but also the cumulative value. I am also concerned about the performance of this solution; it might not be a problem here, but in general we should not recalculate values every time we query them.

PR #1367, opened 2022-05-02
Move PaletteColor to commonMain

kursadsaka:
Moves PaletteColor to commonMain for the https://github.com/iSoron/uhabits/issues/1075

### DIFF
--- a/uhabits-android/src/androidTest/java/org/isoron/uhabits/BaseUserInterfaceTest.kt
+++ b/uhabits-android/src/androidTest/java/org/isoron/uhabits/BaseUserInterfaceTest.kt
@@ -24,8 +24,8 @@ import android.content.Intent
 import androidx.test.core.app.ApplicationProvider
 import androidx.test.platform.app.InstrumentationRegistry
 import androidx.test.uiautomator.UiDevice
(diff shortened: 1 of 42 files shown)


iSoron:
Thanks for the PR, @kursadsaka, but `PaletteColor` is already free from Java dependencies. We should also not move it to `org.isoron.platform.models`; the current package is correct.

kursadsaka:
@iSoron I understand, but I'm asking just to be sure. There is a section in the https://github.com/iSoron/uhabits/issues/1075 for the no direct dependencies files. Should this description be changed or am I misunderstanding?

iSoron:
@kursadsaka Moving files to `commonMain` is welcome, but the files in the "no dependencies" section can just be moved directly; there is no need to change packages, or the contents of these files. For example, `PaletteColor` should still be in `org.isoron.uhabits.core.models`. Please feel free to submit a new PR. There is also no need to submit a separate PR for each file that can be moved directly without changes; one PR for all these files would be better.

kursadsaka:
Thhanks for the explanation @hiqua. I will come back with another PR 😁

PR #1425, opened 2022-07-06
fix marker scaling in frequency display

### LINKED ISSUE #1355: Frequency display does not normalize for the number of times a day of the week occurs in a month
## Pre-submission checklist

## Description
When you do a habit every day, the bubbles in the frequency section are larger for days that happen more times in the month. This makes it look like you're skipping days.

The bubble size should be based on (number of Mondays with habit observed in the month/number of Mondays), not the count of Mondays with the habit observed.

Alternatively, it should show counts instead of bubbles, as smaller bubbles make it seem like days were missed, causing confusion.

## Steps to reproduce
1. Create a new yes/no habit
2. Enter the detailed display for the habit
3. Edit the history to say you did it every day last month (last month cannot have 28 days)
4. See that the frequency bubbles are bigger for the days of the week that happened 5 times last month

## System information
 - Phone: Google Pixel 5
 - Phone Operating System: Android 12
 - App version: 2.0.3

eduebernal:
Fixes #1355 

The marker scaling is now based on the ratio of (number of [Weekday] with habit observed in the month/number of [Weekday]), as proposed in the issue.

This screenshot shows weekdays with 100% completion:
Screenshot_20220705-213212_Habits

This screenshot shows weekdays with varying completion percentages:
Screenshot_20220705-213155_Habits

I have some concerns with the scaling percentages. On weekdays that appear 4 times in a month, the possible percentages are (0,25,50,75,100%), and on weekdays that appear 5 times, the percentages are (0,20,40,60,80,100%). So, for example if you missed the habit 1 time for each weekday, the percentage for some markers would be 75% and for others would be 80%. The difference is barely noticeable but I just wanted to point it out.

Another thing is that this frequency chart may be only meaningful with habits that are done everyday or on specific weekdays.

Tested with a samsung galaxy A71.

Let me know what you think.


uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/utils/DateUtils.kt
```diff
@@ -178,6 +179,26 @@ abstract class DateUtils {
...
+        fun getWeekdaysInMonth(startOfMonth: Timestamp): Array<Int> {
+            val month = startOfMonth.toCalendar()[Calendar.MONTH] + 1
+            val year = startOfMonth.toCalendar()[Calendar.YEAR]
+            val weekday = startOfMonth.weekday
+            val extraWeekdays = YearMonth.of(year, month).lengthOfMonth() - 28
+
+            val freq = Array(7) { 4 }
+            for (day in weekday until weekday + extraWeekdays) {
+                freq[day % 7] = 5
+            }
```
hiqua, line 197: I think it might more readable if you just iterate through all the days of the month and add +1 to the corresponding entry in the array at every iteration.

uhabits-android/src/main/java/org/isoron/uhabits/activities/common/views/FrequencyChart.kt
```diff
@@ -222,15 +224,16 @@ class FrequencyChart : ScrollableChart {
...
-    private fun drawMarker(canvas: Canvas, rect: RectF?, value: Int?) {
+    private fun drawMarker(canvas: Canvas, rect: RectF?, value: Int?, frequency: Int) {
         // value can be negative when the entry is skipped
         val valueCopy = value?.let { max(0, it) }
 
         val padding = rect!!.height() * 0.2f
         // maximal allowed mark radius
         val maxRadius = (rect.height() - 2 * padding) / 2.0f
         // the real mark radius is scaled down by a factor depending on the maximal frequency
-        val scale = 1.0f / maxFreq * valueCopy!!
```
hiqua, line 233: Do we still need `maxFreq`?

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/utils/DateUtils.kt
eduebernal: Yeah I see what you mean, I went with this approach to reduce the time complexity of the algorithm, but I can see how it may be confusing to read.

uhabits-android/src/main/java/org/isoron/uhabits/activities/common/views/FrequencyChart.kt
eduebernal: I don't think so, it was only used in the previous drawMarker method and on an unused method populateWithRandomData. We could also get rid of the getMaxFrequency method.

I just wanted to make sure that you prefer this new approach of showing the percentage rather than the frequency, because this may only be useful for everyday habits were the actual day of the week is meaningful.

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/utils/DateUtils.kt
eduebernal: Changed it to the method you described.

eduebernal:
I fixed the test, I incorrectly assumed that Sunday was 0, but looking at the Timestamp class I noticed that Saturday mapped to 0 instead. 

I couldn't run the tests on my machine because of a mockito error, probably a misconfiguration on my end. But I ran the build & test workflow on my fork and the core tests seems to work fine now.

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/utils/DateUtils.kt
```diff
@@ -178,6 +179,26 @@ abstract class DateUtils {
...
         }
 
+        /**
+         * Returns a vector of Int representing the frequency of each weekday in a given month.
+         *
+         * @param startOfMonth a Timestamp representing the beginning of the month.
+         */
+        @JvmStatic
+        fun getWeekdaysInMonth(startOfMonth: Timestamp): Array<Int> {
+            val month = startOfMonth.toCalendar()[Calendar.MONTH] + 1
```
hiqua, line 189: Why `+1`?

eduebernal: The Calendar class encodes the months with numbers from 0 to 11, but the YearMonth.of(year, month) method used below to get the length of the month expects an int from 1 to 12 for the month, so I just add 1 to use that method correctly.

iSoron:
Thank you for the PR, @eduebernal. I fixed the test screenshots and merged it. For the record, the commands were renamed to `build.sh android-tests <API>` and `build.sh android-accept-images`.

iSoron:
Hi @eduebernal, I just noticed that this PR breaks the frequency chart for numerical habits (see screenshot attached below). If you think you could fix this in the next few days please let me know, and I will wait for your fix before we release 2.1.0. Otherwise, I will revert the commits, and we can re-apply them later.

Screenshot_20220903_201436

eduebernal:
Sure, I submitted a new PR #1489, however if you're not satisfied with that fix, I cannot promise that I will have time to work on it outside of weekends.

PR #1454, opened 2022-08-13
Reimplementing the multiple popups handling in the new popup solution.

kalina559:
The previous solution was done in NumberPickerFactory class, that has been since removed. Moved the changes to the NumberPopup class, that's now used to create numerical popups.


iSoron:
Looks good to me. Let's just wait for the checks to pass. Do we also need this for the Yes/No popup?

kalina559:
Oh yeah, I forgot that this changed as well. Previously when double clicking yes/no it just changed between yes/skipped/no and there was no popup.

kalina559:
I gave up with the inheritance in the end, because the dialog classes sometimes inherit from AppCompatDialogFragment, sometimes from AlertDialog and sometimes from nothing at all. They use different methods to be created/dismissed so it was hard to create a universal way to handle it.

kalina559:
Maybe I'll look at it as part of the next release to refactor the dialogs and align them a bit?

iSoron:
Hi @kalina559, thanks for working on this. A few suggestions: 

1. I think we can make `currentDialog` a top level variable; we don't gain much by putting it inside `HabitsApplication`; it's still just a global variable.
2. If we make it a `WeakReference<Dialog>`, then we don't need to worry about setting it to null.
3. We can use extension functions to make using it more natural.

In summary, I think we could define, in a new file...

```kotlin
var currentDialog: WeakReference<Dialog> = WeakReference(null)

fun Dialog.dismissCurrentAndShow() {
    currentDialog.get()?.dismiss()
    currentDialog = WeakReference(this)
    show()
}
```

...then call `dialog.dismissCurrentAndShow()` instead of `dialog.show()`. Please let me know if I am missing anything.

kalina559:
Something like this? My only worry is that after opening and closing a dialog, it stays saved as the currentDialog, and when opening the next one, we try to dismiss a dialog that's already been closed. No error or warning is thrown though.

kalina559:
Neverming, this doesn't work yet 😀

kalina559:
Okay, now it works. The downside is the delay between showing the the first dialog, closing it and opening the next one after - previously we closed immediately in show(), onCreate() or onCreateDialog() and now we create both dialogs and then check if the second one is not a 'duplicate'. 
But on the other hand this is pretty much an edge case - no one in their right mind opens these dialogs that fast xd
I also found another one of these dialogs in EditHabitActivity.

uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/ColorPickerDialog.kt
```diff
@@ -28,8 +31,14 @@ import org.isoron.uhabits.utils.toPaletteColor
 class ColorPickerDialog : ColorPickerDialog() {
     fun setListener(callback: OnColorPickedCallback) {
         super.setOnColorSelectedListener { c: Int ->
-            val pc = c.toPaletteColor(context!!)
+            val pc = c.toPaletteColor(requireContext())
```
iSoron, line 34: This is a nice refactoring, but could you move it to a separate PR?

uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/MultipleDialogsHandler.kt
```diff
@@ -0,0 +1,15 @@
+package org.isoron.uhabits.activities.common.dialogs
+
+import android.app.Dialog
+import java.lang.ref.WeakReference
+
+class MultipleDialogsHandler {
```
iSoron, line 6: This class is redundant. You can remove it, then move the companion object code to the top level:

```kotlin
import android.app.Dialog [...]

var currentDialog: ...

fun Dialog.dismissCurrent() { ...}
```
```

File: uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/MultipleDialogsHandler.kt [hunk 3]
```diff
@@ -0,0 +1,15 @@
...
+    companion object {
+        var currentDialog: WeakReference<Dialog> = WeakReference(null)
+
+        fun Dialog.dismissCurrent() {
+            currentDialog.get()?.dismiss()
+            currentDialog = WeakReference(this)
+        }
```
iSoron, line 13: Splitting my suggested `dismissCurrentAndShow` function into multiple functions is fine, but it's unexpected to me that `dismissCurrent` also sets the reference. I think `dismissCurrent` should just dismiss and (possibly) set the current reference to null. It would be more clear to have a separate `Dialog.setCurrent()` function, or a combined `showAndSetCurrent()`.

uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/WeekdayPickerDialog.kt
```diff
@@ -73,7 +74,10 @@ class WeekdayPickerDialog :
             .setNegativeButton(
                 android.R.string.cancel
             ) { _: DialogInterface?, _: Int -> dismiss() }
-        return builder.create()
+
+        val dialog = builder.create()
+        dialog.dismissCurrent()
+        return dialog
```
iSoron, line 80: After we return this dialog, does the other class call `dialog.show()`? If so, I think we should move `dismissCurrent` there.

iSoron (changes requested):

uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/WeekdayPickerDialog.kt
kalina559: In this case, DialogFragment.show() is called, so I don't really have access to the dialog until onCreateDialog() of this AppCompatDialogFragment is finished. That's why I put this code here.

kalina559: I figured it out in the end, but it required making separate methods for handling a regular Dialog and DialogFragment

kalina559:
Is there any way to check the ktlint report from Github Actions? Because I have no idea what's causing the checks to fail

PR #1458, opened 2022-08-14
Replaced activity!! with requireActivity, context!! with requireContext etc.

kalina559:
Replaced activity!! with requireActivity, context!! with requireContext etc.


iSoron:
Looks good. Thanks, @kalina559!

PR #1489, opened 2022-09-04
fix marker scaling for numerical habits in frequency display

eduebernal:
Follow up to: #1425

The scaling was broken because inside the FrequencyChart class, the freq HashMap for numerical habits contains the achieved target for each day, so each value could be greater than 1. So the logic to normalize the scaling for Yes or No habits did not apply here.

I thought the best solution would be to restore the previous scaling logic, where the maximum frequency is used to scale the markers, only for numerical habits. That way you can visualize the days when you achieved a higher number, even if that number surpasses your target. If you only want to know if you reached your target, the calendar display does a great job already. 

Summarizing, for numerical habits, the frequency display allows you to visualize the actual values you reached, and the calendar display allows you to visualize if you reached your target or not.


iSoron:
Thanks @eduebernal, that works.

Issue #172, opened 2016-09-11
Persist notifications across reboots and upgrades


iSoron:
By default, when the phone is rebooted or when the application is destroyed for any reason (for example, when upgrading), all the notifications are lost. It would be nice if the app could restore the previous notifications after starting up.

This was originally suggested in #106.

**Implementation:** The app already knows which notifications are active at any time. This information is available in the hash map `NotificationTray.active`. This hash map could be serialized and stored in a SharedPreference, then restored at startup.

DanielKaparunakis:
I would like to work on this. I was trying to see if I could replicate the issue but I was not able to. I set a reminder, turned the phone off, then turned it back on, waited for the reminder to show up and it did. Is there something I am missing?

woj-tek:
OK, steps to reproduce it:
* create a habit (let's say it's 10:00am)
* set notification for 10:05am
* wait for the notification to show up
* when the notification is shown, reboot your phone

expected result: notification should be displayed after phone finished booting up

what actually happens: there is no notification.

--

Haven't tested recently if snoozing works across reboots (set notification, notification is displayed, snooze it, rebook your phone, after reboot and snooze time passed notification should be displayed) but I'm fairly certain that it wasn't working previously as well.

DanielKaparunakis:
Okay I was able to reproduce this. I am thinking that a possible solution is storing the HashMap and then accessing it during device startup through a service launched after a receiver receives the boot complete event. Thoughts?

iSoron:
I think that would work fine for persisting notifications across reboots, but probably not after upgrades. Perhaps we could restore the notifications [when the application is created][1]? This would also restore them after the app is forced to stop then launched again.

[1]: https://github.com/iSoron/uhabits/blob/638b3f763cda02dd8d77d1ff0f456ba733d6790e/app/src/main/java/org/isoron/uhabits/HabitsApplication.java#L111-L112

iSoron:
I no longer think this is an important feature. This affects all Android apps, and should be handled by the operating system. There also seems to be very low user demand for this feature.

digitalcircuit:
Some other applications, such as GMail, do recreate notifications after reboot and application upgrade.  I can understand not prioritizing this, however!

And while not ideal, this encourages me to finish my habits if I need to reboot my phone for other reasons (updates, Magisk module changes, etc) :slightly_smiling_face:

felixwiemuth:
[...]

PR #1509, opened 2022-09-17
Restore notifications when app starts

felixwiemuth:
### Overview

This pull request makes notifications from habit reminders survive reboots, app updates and force-closing by recreating them on app startup. After reboot and app update, notifications reappear without having to start the app manually, after a force-close, they reappear after starting the app manually. In all of these cases, notifications were previously gone and not recreated.

When notifications that had been shown before are recreated, they are shown silently, to not give the impression that they just became due. As before, notifications display the original reminder time - that is, when a notification is recreated, it doesn't display as "now", but shows the relative time since the reminder was due.

This implements #172.

### Implementation
Technically, this is implemented in the following way:

- Active notifications are persisted whenever they are changed, and reloaded from persistent storage whenever the `NotificationTray` class is instantiated.
  - Persistence is achieved through serializing the map with `NotificationData` objects (which now also includes a flag whether the notification has been shown already) to `SharedPreferences`, whenever this map is changed.
  - With persistency, it becomes more important to also remove no more active notifications from the map of active notifications, as they otherwise would be restored, but not reshown, and thus accumulate.
- Whenever the app is created, `HabitsApplication` calls `NotificationTray.reshowAll()` which results in all active notifications being reshown. This also happens after reboot (because of receiving the `BOOT_COMPLETED` intent) and after app upgrade (because of receiving the `MY_PACKAGE_REPLACED` intent).
- Previously, `ReminderReceiver` initiated a rescheduling directly; this was and is still not necessary, as it was and is initiated in `HabitsApplication`.

### Testing
Manually tested successfully the following on a Pixel 4a (Android 13) to check whether it works as intended, including some previous behaviour:

- Force stop app while notification showing, it shows again silently when opening app again
- Force stop app before notification becomes due
  - Start app again before notification becomes due, it shows normally when due
  - Do not start app: notifications are not shown when due (this should be expected, as alarms are cancelled by force-stopping)
    - **However, notifications due (but not shown previously) should be shown when eventually starting the app, but currently don't - this problem already exists in the current release version (2.1.0).**
- Force stop app after having marked habit as done in notification, it does not show again when starting app
- Reboot while notification showing, it shows again silently about 2min after login (this is normal, it can be delayed a bit depending on when Android sends the reboot intent to the app)
- Reboot before notification is due (while it is not showing), login after it is due: **does not show - this problem also exists in release version 2.1.0, as above for the force-stop case**
- Updating app: active notifications show up silently after installing the update without manually starting the app
- Changing reminder time while notification is showing: notification stays, but renews with sound etc. at the new due time
  - The same with rebooting (or force-stopping) before it becomes due and it becoming due before logging in (or starting the app, respectively): shows silently with old due time (NOTE: I would say this is **not optimal** but acceptable; I guess this is due to the same reason that new notifications are not shown when they became due while the app was not running, see above)
    - The same but with logging in / receiving boot intent before it becomes due: it first shows silently with old due time and then renews with sound at the new due time
- When a notification is still active from the previous day, it gets replaced by the new one from the current day when it becomes due


felixwiemuth:
Note that I changed the implementation so that notifications are first added to the active map when the `ShowNotificationTask` is executed, and not when this task is created and scheduled to execute. I guess that this is fine and it is very unlikely that the app stops before the task is run?

This also means that we actually don't have to persist the `shown` field of `NotificationData`, as all notifications added to the active map have been shown (or are just about to be shown). I am currently testing this variant (it is a small change) and I think this will be preferable, except anyone sees an advantage of the current version.

hiqua:
Thanks a lot! Any way to have automated tests for this?

felixwiemuth:
> Any way to have automated tests for this?

I looked into the existing unit tests but didn't find anything related, e.g. some stubbing/test setup that would allow to test whether persistence works correctly.

Regarding instrumented tests, I am not sure how straight-forward it is to test the relevant situations (reboot, force-stop, update). What could be useful is to test whether after starting the app, previously serialized active notifications are shown (however, one has to make sure that they actually should be shown, there are some conditions). At best the serialized notifications are created through the UI as well (schedule habit reminder so that it becomes due). But I am not enough into instrumented tests to say how easy this is.

felixwiemuth:
I have now also tested that a notification from the previous day gets replaced by the new one from the current day when it is due (in the slightly updated version without persisting `shown` though).

felixwiemuth:
I now also pushed the small update with removing the unnecessary `shown` field from `NotificationData`. I executed all manual test from above again and they still seem to pass in the same way. However, I updated, corrected and extended the test list a bit, see in bold for some undesirable behaviour.

I think what should be fixed is that reminders becoming due while the device is off are not eventually shown when the device or even the app is started. This exists, as noted, already in the current release version 2.1.0.

If you agree that this should be fixed, I would take a look.

uhabits-android/src/main/java/org/isoron/uhabits/receivers/ReminderReceiver.kt
```diff
@@ -96,7 +96,7 @@ class ReminderReceiver : BroadcastReceiver() {
                 }
                 Intent.ACTION_BOOT_COMPLETED -> {
                     Log.d("ReminderReceiver", "onBootCompleted")
-                    reminderController.onBootCompleted()
+                    // NOTE: Some activity is executed after boot through HabitsApplication, so receiving ACTION_BOOT_COMPLETED is essential.
                 }
```
iSoron, line 100: Could you clarify why do we still need this block? We would still receive `ACTION_BOOT_COMPLETED`, even if we remove the block.

uhabits-android/src/main/java/org/isoron/uhabits/receivers/UpdateReceiver.kt
```diff
@@ -0,0 +1,14 @@
...
+import android.content.Intent
+import android.util.Log
+
+class UpdateReceiver : BroadcastReceiver() {
+
+    override fun onReceive(context: Context, intent: Intent) {
+        // Dummy receiver, relevant code is executed through HabitsApplication.
+        Log.d("UpdateReceiver", "Update receiver called.")
+    }
+}
```
iSoron, line 14: Why do we need a new (empty) receiver? Could we add `MY_PACKAGE_REPLACED` to the list of intent filters of ReminderReceiver instead?

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/ui/NotificationTray.kt
```diff
@@ -83,9 +105,9 @@ class NotificationTray @Inject constructor(
         return (id % Int.MAX_VALUE).toInt()
     }
 
-    private fun reshowAll() {
+    fun reshowAll() {
         for ((habit, data) in active.entries) {
-            taskRunner.execute(ShowNotificationTask(habit, data))
+            taskRunner.execute(ShowNotificationTask(habit, data, true))
```
iSoron, line 110: For boolean arguments, it's best to add the argument name (`shown = true`)

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/ui/NotificationTray.kt
```diff
@@ -95,18 +117,26 @@ class NotificationTray @Inject constructor(
...
+    @Serializable
+    internal class NotificationData(
+        val timestamp: Timestamp,
+        val reminderTime: Long,
+    )
+
+    private inner class ShowNotificationTask(
+        private val habit: Habit,
+        private val data: NotificationData,
+        private val shown: Boolean
```
iSoron, line 136: `silent` is a bit more clear to me. Changing it would also make it consistent with the other class.

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/preferences/Preferences.kt
```diff
@@ -135,6 +141,23 @@ open class Preferences(private val storage: Storage) {
...
+        val serialized = Json.encodeToString(activeById)
+        storage.putString("pref_active_notifications", serialized)
+    }
+
+    internal fun getActiveNotifications(habitList: HabitList): HashMap<Habit, NotificationTray.NotificationData> {
+        val serialized = storage.getString("pref_active_notifications", "")
+        return if (serialized == "") {
+            HashMap()
+        } else {
+            val activeById = Json.decodeFromString(MapSerializer(Long.serializer(), NotificationTray.NotificationData.serializer()), serialized)
```
iSoron, line 155: What happens if the decoding fails? Does the app fail to boot? I think this could happen if: (i) user receives some notifications; (ii) we update `NotificationData` with some new/renamed fields; (iii) user installs updates and restarts the app. The new version of the app would not be able to read serialized data from the old version.

I suggest adding a try/catch block here, and returning `HashMap()` if the operation fails.

iSoron (changes requested):
Hi @felixwiemuth, thank you for the pull request. I tried it out, and it seems to work well. Please see some comments below.

Regarding testing, I suggest not testing any `uhabits-android` classes at this time, but I think it's important to write tests for all changes in `uhabits-core`. For example, we could:

1. Store notifications to preferences then immediately retrieve them and check that the retrieved version equals the original one.
2. Create a NotificationTray with a mock Preferences, and ensure that the correct calls to `preferences.setActiveNotifications` are made. Also ensure that `set/remove/get` operations behave as expected.

iSoron:
> I think what should be fixed is that reminders becoming due while the device is off are not eventually shown when the device or even the app is started. This exists, as noted, already in the current release version 2.1.0.

This would be a nice feature, but it could also be a separate PR.

uhabits-android/src/main/java/org/isoron/uhabits/receivers/ReminderReceiver.kt
felixwiemuth: The block is not needed (and I can see the intent is also logged above); What I deemed important is that somewhere it is noted that it is important that the `ACTION_BOOT_COMPLETED` intent is received.

Actually, as for both `ACTION_BOOT_COMPLETED` and `MY_PACKAGE_REPLACED` the only thing we need is that the application is started (and all the code from `ReminderReceiver` is not needed), I would suggest that both should be received by a common dummy receiver (as currently with `UpdateReceiver`), which we could call `StartAppReceiver` or similar. Then it is clear that these intents are solely received to let the app start.

uhabits-android/src/main/java/org/isoron/uhabits/receivers/UpdateReceiver.kt
felixwiemuth: See comment above.

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/ui/NotificationTray.kt
felixwiemuth: Done.

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/ui/NotificationTray.kt
felixwiemuth: Agree, done.

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/preferences/Preferences.kt
felixwiemuth: Yes currently there would be an uncaught exception if decoding fails, and the app would crash on startup. That is of course bad, even though unlikely. Returning an empty map in case it cannot be deserialized is fine, as (as far as I can see) no other functionality depends on all active notifications being available in the map.

What we could do is show a notification that notifications could not be restored and the user should make sure to look at the habits manually, but regarding that it is a corner case it's probably not worth it.

Btw., decoding will not fail if fields not present in the encoded string have default values in the Kotlin class.

felixwiemuth:
> > I think what should be fixed is that reminders becoming due while the device is off are not eventually shown when the device or even the app is started. This exists, as noted, already in the current release version 2.1.0.
> 
> This would be a nice feature, but it could also be a separate PR.

I agree that this should be a separate PR (or issue for now). I have also a few other related things where I'd open a new issue for.

felixwiemuth:
> Regarding testing, I suggest not testing any `uhabits-android` classes at this time, but I think it's important to write tests for all changes in `uhabits-core`. For example, we could:
> 
>     1. Store notifications to preferences then immediately retrieve them and check that the retrieved version equals the original one.

I'm currently looking into adding a test to `PreferencesTest` but am looking for some help with map comparison (see https://stackoverflow.com/questions/73852632/test-equality-of-collections-with-custom-matcher).

> 
>     2. Create a NotificationTray with a mock Preferences, and ensure that the correct calls to `preferences.setActiveNotifications` are made. Also ensure that `set/remove/get` operations behave as expected.

I started writing a test with a `MockPreferences : Preferences(DummyStorage())` class which simply retains the map passed to it via `setActiveNotifications`. For this I need to make the set and get operations `open`, so that they can be overridden with the mock functionality.
Then we can test the public methods `show` and `cancel` of `NotificationTray`, and potentially the four special cases where habits should be removed from the active list during execution of `ShowNotificationTask`.

uhabits-core/src/jvmTest/java/org/isoron/uhabits/core/ui/NotificationTrayTest.kt
```diff
@@ -0,0 +1,125 @@
...
+    fun testShow() {
+        // Show a reminder for a habit
+        val habit = fixtures.createEmptyHabit()
+        habit.reminder = Reminder(8, 30, WeekdayList.EVERY_DAY)
+        val timestamp = Timestamp(System.currentTimeMillis())
+        val reminderTime = System.currentTimeMillis()
+        notificationTray.show(habit, timestamp, reminderTime)
+
+        // Verify that the active notifications include exactly the one shown reminder
+        // TODO are we guaranteed that task has executed?
```
felixwiemuth, line 112: As a `ShowNotificationTask` is executed by a task runner, I am not sure whether this might be asynchronous and we should for example add a delay here to increase the likelihood that it has executed? Or are tasks (at least for these tests) run in the same thread, which would make tests safer?

uhabits-core/src/jvmTest/java/org/isoron/uhabits/core/ui/NotificationTrayTest.kt
```diff
@@ -0,0 +1,125 @@
...
+        assertThat(
+            preferences.getActiveNotifications(habitList)[habit],
+            equalTo(NotificationTray.NotificationData(timestamp, reminderTime))
+        )
+
+        // Remove the reminder from the notification tray and verify that active notifications are empty
+        notificationTray.cancel(habit)
+        assertThat(preferences.getActiveNotifications(habitList).size, equalTo(0))
+
+        // TODO test cases where reminders should be removed (e.g. reshowAll)
```
felixwiemuth, line 123: Here we could add tests whether the remaining methods which touch the active notifications map remove entries in the correct situations. It is a little work to setup all the special cases, I can't promise right now when I'll be able to do that.

uhabits-core/src/jvmTest/java/org/isoron/uhabits/core/preferences/PreferencesTest.kt
```diff
@@ -162,6 +165,31 @@ class PreferencesTest : BaseUnitTest() {
...
+            a[habit] = data
+        }
+
+        // Persist and retrieve active notifications
+        prefs.setActiveNotifications(a)
+        val b = prefs.getActiveNotifications(habitList)
+
+        // Assert that persisted and retrieved maps are teh same
+        assertThat(a.keys, equalTo(b.keys))
+        a.forEach { e -> assertThat(b[e.key], equalTo(e.value)) }
```
felixwiemuth, line 190: This is a manual workaround for the missing matcher/library function I was looking for.

felixwiemuth (commented):
I have now added the described tests in their current state. They check most of what we discussed. See the comments for further info.

hiqua:
@iSoron if the state of the PR seems ok I can try to rebase it, let me know. AFAICT from reading the comments, all the concerns were addressed.

BURG3R5:
This is a great QoL feature so can we please get this merged?

PR #1514, opened 2022-09-25
adding support for 1x2 widget size for history widget

pphu:
This adds support for 1x2 history widget size by using Java Calendar's [NARROW_FORMAT](https://docs.oracle.com/javase/8/docs/api/java/util/Calendar.html#NARROW_FORMAT) when getting narrow. From discussion #1513  
1x2 widget example

I've been removed from Android development for many years but from what I can tell, Android 13 has better support for widget size. However since the min SDK is Android 9, I'm using the `width < 80` to determine if the widget size is less than 2 cells wide. The default widget size is currently at 80 and that rounds to 2 cells wide. It's a little too magical for my taste but if anyone has a suggestion, please let me know.


hiqua:
> Android 13 has better support for widget size. However since the min SDK is Android 9, I'm using the width < 80 to determine if the widget size is less than 2 cells wide.

If that's the case, can you branch on the Android version depending on that, such that the magic only happens for Android 12 and lesser? This way we can remove the magic eventually, once we upgrade the SDK.

iSoron:
Thanks for the PR, @pphu. The implementation looks good to me, but I think we need unit tests for `narrowWeekdayName` and a view test for HistoryChart (under the new size). See [HistoryCardViewTest.kt](https://github.com/iSoron/uhabits/blob/d95084500ffef7b45e3a7d7e655d5d9c7a543a30/uhabits-android/src/androidTest/java/org/isoron/uhabits/activities/habits/show/views/HistoryCardViewTest.kt) for a view test example and [TEST.md](https://github.com/iSoron/uhabits/blob/dev/docs/TEST.md) for instructions on running view tests.

Issue #1532, opened 2022-10-04
Comma error when using app in Italian


Edgiest05:
## Pre-submission checklist
- [x] I am submitting a bug report, not a feature request.
- [x] I have searched for similar issues, but did not find any matches.

## Description
When using the app in Italian, values are stored using the comma "," as separator for decimal units.
When modifying an entry (with decimals) the old value is autofilled inside the prompt using the comma that makes new values fail when inserted.

## Steps to reproduce
1. Go to app's home page (using Italian as language)
1. Click on value with decimals to modify it
1. Keep the comma as suggested but change the value
1. See the value not changing

## System information
 - Phone: [Xiaomi Redmi Note 11 Pro 5g]
 - Phone Operating System: [Android 11]
 - App version: [2.1.1]

iSoron:
Thanks for the detailed bug report, @Edgiest05. I was able to reproduce it.

Edgiest05:
Glad it helped.
Amazing work, keep it up!

iSoron:
Fixed in 0bb82a4

PR #1548, opened 2022-10-13
Fixed Italian decimals issue

Edgiest05:
This commit fixes iSoron/uhabits#1532
Comma replaced with dot when reading from dialog input


iSoron:
Thank you for the PR, @Edgiest05. Unfortunately the proper fix is more complicated than replacing the comma. See 0bb82a4

Edgiest05:
Mine was a really cheeky fix indeed. Anyway, thank you @iSoron for your time.

PR #1497, opened 2022-09-07
Support for Android 13 Themed Icons 

cheeeeer:

PR #1554, opened 2022-10-18
Implement per-app language setting for Android 13

leondzn:
# Overview

This addresses #929 and #879 

Android 13 introduces per-app language setting where you can now change the app language on a centralized location in the system settings without relying on the currently active system language. And also without having to create a custom language picker inside the app.

Accessible via `Settings > System > Languages & Input > App Languages`

However only apps that explicitly support Per-App Language settings will appear on the menu.

This PR enables that.

There are two changes that have been done in order to do that:

1. Updating the compile SDK version to API level 33 (Android 13)
    This will enable us to use the `android:localeConfig` attribute on the AndroidManifest. I've also took the liberty of also updating the target SDK version to match. In doing so, however, there are a breaking change with the `GestureDetector.OnGestureListener` interface. In which the `MotionEvent` parameters on the interface methods are now non-nullable. There are only two implementing classes namely `AndroidDataView.kt` and `ScrollableChart.kt` in which I updated the implementing method parameters. UI tests that use these classes has also been updated.

2. Added a `locales_config.xml` resource file to specify the supported languages that should appear on the menu. As of now I based the language codes from the available `strings.xml` localized files.

# Testing

I've manually tested this on a Pixel 6 emulator and a physical Pixel 4a both running Android 13.
There is a warning on the Android manifest file about `android:localeConfig` being only available for API 33 and not on lower sdk versions but the app still works fine. I tested on an emulated Pixel 4 running Android 11. My guess is since Per-App language setting doesn't exist in < API 33, the `android:localeConfig` attribute will never be queried anyway.

For automated tests, all existing unit tests passed.
Some of the UI tests are failing even before I made these changes. While tests involving `AndroidDataView.kt` and `ScrollableChart.kt` are still passing.


iSoron:
Thanks for the PR, @leondzn. All the changes look good to me. Android 13 also introduces [Notification runtime permissions](https://developer.android.com/develop/ui/views/notifications/notification-permission). Do we need any changes to support that?

oscfdezdz:
0e1399724ff0bdb5e126ce3629b6419819dd0938 enables monochrome icon introduced in #1497 for me, with the latest stable version it is not displayed.

hiqua:
@leondzn did you have time to have a look?

hiqua:
Superseded by https://github.com/iSoron/uhabits/pull/1783

Issue #1589, opened 2022-12-09
"Rate this app on Google Play" would cause crash if it fails to open play store


Terrible-Tornado:
## Pre-submission checklist
- [x] I am submitting a bug report, not a feature request.
- [x] I have searched for similar issues, but did not find any matches.

## Description
"Rate this app on Google Play" would make the app crash if it fails to open the play store

## Steps to reproduce
1. Go to Settings
2. Scroll down to the bottom
3. Click "Rate this app on Google Play" (This phone don't have play store at this time), then the app crashes

## System information
 - Phone: Google Pixel 2
 - Phone Operating System: Android 9.0
 - App version: 2.1.1

## Screenshots
https://user-images.githubusercontent.com/119578731/206694593-edc8b376-18a2-4a41-840d-8b5687292242.mp4

PR #1590, opened 2022-12-10
Issue 1589: Fixed crash after choosing Settings/Rate app when there's no Play Store on the device

kalina559:
Solves #1589.
Launching Play Store through startActivitySafely (just like it's done in AboutScreen) to handle the exception thrown where Play Store is not installed on the device.


hiqua:
I'm not familiar with the semantics of the return value, but shouldn't we return true as well in the case of `rateApp`?

kalina559:
@hiqua you're right, in this case it will still work if I don't return anything, but returning true means we do a few operations less. It's described [here](https://stackoverflow.com/questions/16604567/preferencefragmentonpreferencetreeclick-return-value-what-does-it-do)

hiqua (approved):

Issue #1576, opened 2022-11-23
Habitbull import numerical value habit bug


Rslayl:
## Pre-submission checklist
- [x] I am submitting a bug report, not a feature request.
- [x] I have searched for similar issues, but did not find any matches.

## Description
When importing Habitbull CSV with numerical value habit, values are off by 3 decimals. Ex: 80 becomes 0.08 in Loop.

## Steps to reproduce
1. Settings > import data > choose habitbull csv
2. Input value for today
3. All data for habit appear
4. Values are off by 3 decimals

## System information
 - Phone: Sony Xperia 5 III
 - Phone Operating System: Android 12
 - App version: 2.1.1

hiqua:
Can you share the csv, redacted if needed?

Rslayl:
[HabitBullData.csv](https://github.com/iSoron/uhabits/files/10079848/HabitBullData.csv)

iSoron:
Fixed in #1591

PR #1591, opened 2022-12-11
Issue 1576: Solving problems with habitbull import

kalina559:
Solves #1576.

Habitbulls values are always 1000 times smaller than ours, that's why we need to multiply it during the initial import.
I also noticed that the values aren't visible after import but before inputting the first value in Loop Habit Tracker, that's why I invoked Habit.recompute() right after the import.


hiqua:
Can you add a test using the input from the issue? https://github.com/iSoron/uhabits/issues/1576#issuecomment-1325783183

kalina559:
Sure, I also had to correct testHabitBullCSV3 because there were some numerical values there as well.

iSoron:
Looks good to me. Thanks, @kalina559.

Issue #1574, opened 2022-11-18
Custom snooze period in dark system theme > watch hand is hardly visible


uli-on:
## Pre-submission checklist
- [x] I am submitting a bug report, not a feature request.
- [x] I have searched for similar issues, but did not find any matches.

## Description
When the system is set to a dark theme the watch hands have almost zero colour contrast with the background clock-face, independent of the theme chosen inside the app.

## Steps to reproduce
1. Set the system to a dark theme
2. Snooze an event from its notification
3. Choose a custom period
4. Try to spot the watch hands

## System information
 - Phone: fairphone 3
 - Android 10
 - App version: 2.1.0

## Screenshots
Can't upload images in Octodroid, the app I'm using for participating in GH discussions.

uli-on:
@kalina559 Thanks for all this, Jakub, I think your screenshot brought us closer: please try to replicate with another habit colour, the lightest neutral grey, third from last. Then you should get this one here:

![Screenshot_20221211-170244](https://user-images.githubusercontent.com/35634829/206915918-82820fa7-b30f-415b-990d-67c162c77c93.png)

kalina559:
Okay now I was able to reproduce this. I don't think the bug is very serious, as the user is still able to see the time as digits, but it's a bit weird that the watch hands have different color than when you edit the reminder time in app 😄

uli-on:
Yeah, it's a bit odd, especially cause I chose that colour for its best
visibility

uli-on:
Thanks!

PR #1592, opened 2022-12-11
Issue 1574: Fixed the almost invisible watch hands in snooze activity

kalina559:
Solves #1574.

The bug was caused by the incorrect setting of themeSwitcher.currentTheme (or lack thereof). So in SnoozeDelayPickerActivity line 64, the androidColor was always set using LightTheme, even if we ran activity.setTheme(R.style.BaseDialogDark), hence the problem described in the issue.

Initially I wanted to solve it by running themeSwitcher.apply(), just like it's done in EditHabitActivity, but it seems we can't use R.style.AppBaseThemeDark from the SnoozieDelayPickerActivity - the dialog was turning completely white when I tried that.


iSoron:
Thanks, @kalina559. Do we need to apply a similar fix in any other dialogs?

kalina559:
I didn't really check other dialogs yet. We may use the wrong theme there as well, but probably the dialogs are pretty readable anyway. I'll look into it and create a new issue if that's the case.

PR #1666, opened 2023-03-19
Added content description for icon image in about section. 

vikasofvikas:


iSoron:
Thanks for the PR @vikasofvikas, but the content description you provided does not really improve the app accessibility.

PR #1668, opened 2023-03-25
Update Play Store and F-Droid badges

AdrianMiozga:
I've updated store badges in README.md as the current ones are outdated and also made them a tad bigger as new graphics have more padding around them.


iSoron:
Thanks, @AdrianMiozga!

PR #1669, opened 2023-03-26
Remove dependency on kotlin stdlib

hiqua:
According to:
https://kotlinlang.org/docs/whatsnew14.html#dependency-on-the-standard-library-added-by-default the explicit dependency is no longer needed.


iSoron:
@hiqua Thanks!

PR #1694, opened 2023-04-27
Add a pointer to the FAQ from the README

dimo414:
It took a few minutes longer for me to find the FAQ than I would have liked. It's a super helpful document so it seems worth improving its discoverability.


iSoron:
Thanks for the PR, @dimo414, but the FAQ is already displayed prominently in the app itself.

dimo414:
Well, it took me a while to find :) I started from the GitHub README and it's hard to discover that way. Is there some harm from adding another pointer?

Issue #1695, opened 2023-04-28
Skips impact the score for non-everyday habits


texnic:
### Discussed in https://github.com/iSoron/uhabits/discussions/1686

I have a "6 times per week" habit. I have reached a 100 % score on it.
Then I skipped 4 days (not missed, but skipped).
In case it matters: I am now at "total" of 218.
Depending on how many days passed since the skipping days, the score changes.
Today is 28 April.
If the skipped days are 25-28 April, the score remains 100 %, as expected.
Skipped 24-27 April, habit done again on 28 April; score 99 %.
Skipped 23-27 April, done again on 27-28 April; score 98 %.
Skipped 22-25 April, done again on 26-28 April; score 98 %.
Skipped 21-24 April, done again on 25-28 April; score 97 %.
And so on. If the skipped days are long in the past, today's score starts increasing again.

The expected behaviour is that the score remains at 100 % in all cases above, since the "skips keep the score unchanged and don't break the streak" (by definition).

PR #1713, opened 2023-05-27
Fixed score when using skips on non-daily habit

Maxet1000:
This fixes the problems with the score raised in issue #1695 where a skip affects the score in non-daily habits, when it shouldn't.

To fix the issue, the part of the recompute function that calculates the rolling sum of completed habits in the last period was changed to take skips into account. A new function that calculates amount of skips in an interval is used to expand the length of that period.

Six tests where added, three to test the new helper function and three for testing the score.


uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/ScoreList.kt
```diff
@@ -61,6 +61,31 @@ class ScoreList {
...
 
+    /**
+     * Returns the number of skips after the offset in the interval used to calculate
+     * the percentage of completed days.
+     *
+     * If skips are found in the interval, it is expanded until the interval has the size of the
+     * sum of the denominator and the number of skips within the interval.
+     */
+    @Synchronized
+    fun getNumberOfSkipsByInterval(
```
hiqua, line 72: Not really knowledgeable about this, but if you want to make it recursive, might as well follow https://kotlinlang.org/docs/functions.html#tail-recursive-functions, what do you think?

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/ScoreList.kt
```diff
@@ -124,8 +149,12 @@ class ScoreList {
                     rollingSum += 1.0
                 }
                 if (offset + denominator < values.size) {
-                    if (values[offset + denominator] == Entry.YES_MANUAL) {
-                        rollingSum -= 1.0
+                    val nbOfSkips =
+                        getNumberOfSkipsByInterval(values, offset, offset + denominator)
+                    if (offset + denominator + nbOfSkips < values.size) {
```
hiqua, line 154: Create a variable for `offset + denominator + nbOfSkips` to make it easier to read?

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/ScoreList.kt
```diff
@@ -61,6 +61,31 @@ class ScoreList {
...
+     * the percentage of completed days.
+     *
+     * If skips are found in the interval, it is expanded until the interval has the size of the
+     * sum of the denominator and the number of skips within the interval.
+     */
+    @Synchronized
+    fun getNumberOfSkipsByInterval(
+        values: IntArray,
+        firstIndexToCheck: Int,
+        lastIndexToCheck: Int
```
hiqua, line 75: This parameter name is misleading, because in practice it's not necessarily the last index checked. Can you try to make it clearer for the reader what this does in which cases?

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/ScoreList.kt
```diff
@@ -61,6 +61,31 @@ class ScoreList {
         return result
     }
 
+    /**
+     * Returns the number of skips after the offset in the interval used to calculate
+     * the percentage of completed days.
+     *
+     * If skips are found in the interval, it is expanded until the interval has the size of the
+     * sum of the denominator and the number of skips within the interval.
```
hiqua, line 69: There's no `denominator` in this function so this doc could be improved to make it understandable on its own, can you try to do that?

uhabits-core/src/jvmTest/java/org/isoron/uhabits/core/models/ScoreListTest.kt
```diff
@@ -145,6 +146,93 @@ class YesNoScoreListTest : BaseScoreListTest() {
         checkScoreValues(expectedValues)
     }
 
+    @Test
+    fun test_getNumberOfSkipsByInterval_NoSkips() {
+        val vars = intArrayOf(-1, -1, -1, -1, 3, 2, 2, -1, 2, 2, -1, 2, 2, -1, 3, 2)
```
hiqua, line 151: Here and below: can you hide the integer behind the enum, i.e. use `SKIP` etc. instead?

hiqua (changes requested):
No change needed in the numerical habit case?

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/ScoreList.kt
Maxet1000: This wouldn't work here as the 'tailrec' modifier requires that the function calling itself is the last operation that function performs, which isn't the case here. I'm not really knowledgeable about this either, so I might be missing something.

Maxet1000:
Does anything else need to be changed or can this be integrated? @hiqua @iSoron

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/ScoreList.kt
hiqua: Yes but you can rewrite so that's the case, i.e. you can change the function so that it's tail-recursive. I think that'd be preferable here.

hiqua:
> Does anything else need to be changed or can this be integrated? @hiqua @iSoron

I think we should fix the potential stack overflow with the recursive function before merging this, either by making the function tail recursive as commented, or just switch to an imperative function.

Maxet1000:
@hiqua I've now made the function tail recursive like requested

hiqua:
Could you also rebase on top of the current dev, to make it possible to run the GH actions? Thanks!

Maxet1000:
I've rebased it! I'm not very familiar with rebasing, so I hope I did it correctly. Please let me know if I should make any more changes.

Maxet1000:
I've updated it (again) and it should be possible to merge now

cyril-marpaud:
Any update regarding this one ? 👀

PR #1715, opened 2023-05-31
Update github actions

Jean-BaptisteC:
Bump checkout to v3 because nodejs 12 is deprecated
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/


iSoron:
Thanks for the PR, @Jean-BaptisteC.

PR #1728, opened 2023-06-05
Add downloads shield to readme

mathijs81:
Add shield to show how popular Loop habits is.


iSoron:
Thanks, but I'm not merging this as I'm concerned about appbrain.com tracking every access to our GitHub page.

mathijs81:
Thanks for your concern.
Tracking is not possible on github markdown pages because GitHub anonymizes and caches all external resources as discussed on https://github.blog/2014-01-28-proxying-user-images/
So AppBrain will only see an occasional fetch by github-camo to refresh the image which will then be served by GitHub itself.

Maybe you want to reconsider adding the shield?

PR #1736, opened 2023-06-12
View Skipped Days In The Bar Chart

Mahmoud-Ibrahim-750:
I noticed that the skipped days appear in the history chart but not in the bar chart so I modified the code adding a piece of code to pass the skipped days data from the database to the view and drawing a bar with a slightly transparent colour -to let it be different from the main column- accordingly.


uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/show/views/BarCardView.kt
```diff
@@ -40,6 +40,7 @@ class BarCardView(context: Context, attrs: AttributeSet) : LinearLayout(context,
         val androidColor = state.theme.color(state.color).toInt()
         binding.chart.view = BarChart(state.theme, JavaLocalDateFormatter(Locale.US)).apply {
             series = mutableListOf(state.entries.map { it.value / 1000.0 })
+            skippedSeries = mutableListOf(state.skippedEntries.map { it.value / 1.0 })
```
Mahmoud-Ibrahim-750, line 43: pass the skipped days to the view

uhabits-core/src/jvmTest/java/org/isoron/uhabits/core/ui/views/BarChartTest.kt
```diff
@@ -34,10 +34,12 @@ class BarChartTest {
     val component = BarChart(theme, fmt)
     private val axis = (0..100).map { today.minus(it) }
     private val series1 = listOf(200.0, 0.0, 150.0, 137.0, 0.0, 0.0, 500.0, 30.0, 100.0, 0.0, 300.0)
+    private val series2 = listOf(10.0, 0.0, 50.0, 0.0, 0.0, 0.0, 100.0, 0.0, 60.0, 0.0, 150.0)
 
     init {
         component.axis = axis
         component.series.add(series1)
+        component.skippedSeries.add(series2)
```
hiqua, line 42: How come the test setup seems to change, but the assertions (the .png) do not? Are they just close enough such that the tests pass anyway, and should they be [updated](https://github.com/iSoron/uhabits/blob/dev/docs/TEST.md) instead? What am I missing?

Mahmoud-Ibrahim-750: After running the test multiple times with different cases, here is a final conclusion:
First, the test mechanism seems to ignore little differences between the expected image and the actual one, so yes, they are close enough such that the test pass as long as the difference value is less than 1. 
Second, I think the test mechanism should be updated to take care of little difference, then according to that update, the expected image should be updated too. We may update it for now until we take care of the test in a different PR.

Here are some of the results I got from the test, feel free to try them on your own (each test case is provided with the difference value and the output image too):
1. the current expected image (the one we may update later):
[[Imgur](https://i.imgur.com/59dDEHg.png)]

2. the current test data (with skipped days):
[[Imgur](https://i.imgur.com/1oPJApQ.png)](current test data with skipped days)
[[Imgur](https://i.imgur.com/8dAKgMb.png)](output)
[[Imgur](https://i.imgur.com/dPUrIMS.png)](difference)

3. the current test data (without skipped days):
[[Imgur](https://i.imgur.com/gw1xmMD.png)](current test data without skipped days)
[[Imgur](https://i.imgur.com/NyBa4Aw.png)](output)
[[Imgur](https://i.imgur.com/u3xCRoI.png)](difference)

4. a modified test data (accepted):
[[Imgur](https://i.imgur.com/2dxzoZm.png)](accepted modified test data)
[[Imgur](https://i.imgur.com/DbUWvT4.png)](output)
[[Imgur](https://i.imgur.com/zt9LtHH.png)](difference)

5. a modified test data (rejected):
[[Imgur](https://i.imgur.com/Oq74hpB.png)](rejected modified test data)
[[Imgur](https://i.imgur.com/rvRDqPu.png)](output)
[[Imgur](https://i.imgur.com/thrNato.png)](difference)

hiqua: I think it's better to update the expected images in the same PR, following https://github.com/iSoron/uhabits/blob/dev/docs/TEST.md#running-instrumented-tests, but you can also wait for @iSoron to chime in if you prefer.

Mahmoud-Ibrahim-750: It's ok to wait. Updating the image takes no time anyway, So there's no need to rush.

Mahmoud-Ibrahim-750: @iSoron, Would you check this PR so we can proceed and finally merge it?

Mahmoud-Ibrahim-750: @hiqua Test image is up-to-date and this PR is ready for a final review.

PR #1743, opened 2023-06-16
introducing confetti animation using konfetti library

gokulk16:
- Added Konfetti library: https://github.com/DanielMartinus/Konfetti

Test cases when confetti will be shown for habit type - boolean(Yes/No):
- Toggling with the short press  
- Toggling inside the checkmark pop-up

Test cases when confetti will be shown for habit type - measurable:
- Adding an entry using the number pop-up. animation shows even when the number doesn't match the target for the day respecting the progress made on the day for the user.

Settings combinations tested with: 
- skip days disabled; toggle with short press disabled
- skip days disabled; toggle with short press enabled
- skip days enabled; toggle with short press disabled
- skip days enabled; toggle with short press enabled

Screen recording of confetti animation:

https://github.com/iSoron/uhabits/assets/8376313/e95ec5db-297e-4f74-8905-1a764840246b

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/NumberDialog.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/NumberDialog.kt
@@ -105,7 +105,7 @@ class NumberDialog : AppCompatDialogFragment() {
         val konfettiView = requireActivity().findViewById<LinearLayout>(R.id.konfettiLayout)
 
         if (value > 0.0) {
-            //To motivate, show confetti even if some value is present
+            // To motivate, show confetti even if some value is present
             showConfetti(konfettiView)
         }
     }
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/NumberDialog.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/NumberDialog.kt
@@ -102,11 +102,11 @@ class NumberDialog : AppCompatDialogFragment() {
         val notes = view.notes.text.toString()
         onToggle(value, notes)
         requireDialog().dismiss()
-        val v = requireActivity().findViewById<LinearLayout>(R.id.konfettiLayout)
+        val konfettiView = requireActivity().findViewById<LinearLayout>(R.id.konfettiLayout)
 
         if (value > 0.0) {
-            showConfetti(v)
-
+            //To motivate, show confetti even if some value is present
+            showConfetti(konfettiView)
         }
     }
 }
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/NumberDialog.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/NumberDialog.kt
@@ -7,6 +7,7 @@ import android.view.KeyEvent
 import android.view.LayoutInflater
 import android.view.MotionEvent
 import android.view.View
+import android.widget.LinearLayout
 import androidx.appcompat.app.AppCompatDialogFragment
 import org.isoron.uhabits.HabitsApplication
 import org.isoron.uhabits.R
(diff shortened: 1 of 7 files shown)


uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/NumberDialog.kt
```diff
@@ -100,5 +102,11 @@ class NumberDialog : AppCompatDialogFragment() {
         val notes = view.notes.text.toString()
         onToggle(value, notes)
         requireDialog().dismiss()
+        val v = requireActivity().findViewById<LinearLayout>(R.id.konfettiLayout)
```
hiqua, line 105: `v` is called `konfettiView` above, consistency?

uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/NumberDialog.kt
```diff
@@ -100,5 +102,11 @@ class NumberDialog : AppCompatDialogFragment() {
...
+
+        if (value > 0.0) {
```
hiqua, line 107: Maybe it should be on completion rather than for any value? Not sure.

uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/CheckmarkButtonView.kt
```diff
@@ -88,25 +89,28 @@ class CheckmarkButtonView(
         setOnLongClickListener(this)
     }
 
-    fun performToggle() {
+    fun performToggle(v: View) {
```
hiqua, line 92: Same remark about this `v` variable, make it more explicit so that it's easier to understand.

hiqua (commented):
Just a superficial look at the code for now.

uhabits-android/src/main/java/org/isoron/uhabits/utils/ViewExtensions.kt
```diff
@@ -101,6 +112,32 @@ fun ViewGroup.buildToolbar(): Toolbar {
     return inflater.inflate(R.layout.toolbar, null) as Toolbar
 }
 
+fun ViewGroup.buildKonfettiView(): View {
+    val inflater = LayoutInflater.from(context)
+    return inflater.inflate(R.layout.konfetti, null) as View
+}
+
+fun showConfetti(view: View) {
+    val viewId = R.id.konfetttiView
```
ezio-melotti, line 121:
```suggestion
    val viewId = R.id.konfettiView
```
Is this a typo or is there a reason to have 3 `t`?

uhabits-android/src/main/res/layout/konfetti.xml
```diff
@@ -0,0 +1,29 @@
...
+    android:id="@+id/konfettiLayout"
+    android:layout_width="wrap_content"
+    android:layout_height="match_parent"
+    android:focusable="false"
+    android:orientation="vertical" >
+
+    <nl.dionsegijn.konfetti.xml.KonfettiView
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:id="@+id/konfetttiView"/>
```
ezio-melotti, line 27:
```suggestion
        android:id="@+id/konfettiView"/>
```

gokulk16 (commented):
fixed few comments, fixed conflicts and responded to some comments

hiqua:
1. I think we'd need a new setting to allow users to disable this if they don't want this
2. what happens if you interact with the UI while the animation is working, does it still work? Does it interrupt the animation?
3. maybe something a bit more discrete would be better

I don't want to tell you to do 1. yet because I'm still not sure @iSoron is open to this new UI feature, better to wait for a confirmation first.

neube3:
Just wanted to chip in that the lack of a strong blue colour (e.g. 0000FF) in that confetti would make me not use it. It's not confetti without all primary colours. (Only half-joking)

iSoron:
Thank you for the PR, @gokulk16. I have merged it to dev (06090e238a6f9dc6887fc01d920f252fe2c0e7c0) with the following major changes:

- Move the Konfetti code to ListHabitsScreen
- Make the confetti originate from the button, instead of the center of the screen
- Use the color of the habit as the base color for the confetti, but add some randomization, so we still get multiple colors
- Tweak some animations, layout and the conditions in which the confetti is shown.

Feel free to submit other PRs to tweak it further or fix any remaining issues. At this point, I think the animation is subtle enough for it to be enable by default. If users complain about it, we can add an option in the settings to disable it, or even make it opt-in.

gokulk16:
@iSoron Thanks for the tweaks and also merging the PR. 🙏

PR #1763, opened 2023-07-19
Fixed Use of hardcoded US locale for BarChart

Slartibartfass2:
The selected locale (here `DE`) is not used in the BarChart, but instead `US` is used:
BarChart_DE_bug

When replacing the hard-coded `Locale.US` with `Locale.getDefault()`, the correct short month names are shown:
BarChart_DE_fixed


hiqua:
This looks reasonable, I don't think we have a test covering that and I don't think adding one is necessary, so I'll merge that in a week or after unless told otherwise @iSoron.

PR #1777, opened 2023-07-26
Migrate to view bindings

hiqua:


uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/FrequencyPickerDialog.kt
```diff
@@ -22,97 +22,101 @@ package org.isoron.uhabits.activities.common.dialogs
...
-import kotlinx.android.synthetic.main.frequency_picker_dialog.view.*
 import org.isoron.uhabits.R
+import org.isoron.uhabits.databinding.FrequencyPickerDialogBinding
 
 class FrequencyPickerDialog(
     var freqNumerator: Int,
     var freqDenominator: Int
 ) : AppCompatDialogFragment() {
+    private var _binding: FrequencyPickerDialogBinding? = null
+    private val binding get() = _binding!!
```
iSoron, line 39: Could this be replaced by `private lateinit var binding: FrequencyPickerDialogBinding` ?

hiqua: From https://developer.android.com/topic/libraries/view-binding#fragments, the example is different from the one for activities, I think because of this note further down:

"Note: Fragments outlive their views. Make sure you clean up any references to the binding class instance in the fragment's [onDestroyView()](https://developer.android.com/reference/kotlin/androidx/fragment/app/Fragment#ondestroyview) method."

And `FrequencyPickerDialoger` is a fragment to this should apply to it. So we should be able to assign to `binding` (`_binding` in the current) twice. Or am I missing something?

hiqua:
@iSoron I'll merge this in a week unless told otherwise, I think it's relatively safe given the that it's only about internals and the tests pass.

iSoron:
@hiqua Thanks for the PR. It looks good to me. Please feel free to merge.

Issue #1779, opened 2023-07-28
Cannot view the month name when scrolling through the list horizontally


ninadvadujkar:
## Pre-submission checklist
- [-] I am submitting a bug report, not a feature request.
- [-] I have searched for similar issues, but did not find any matches.

## Description
When scrolling through the days horizontally, only the day and date is displayed. Month information is not provided. Users feel lost and not able to understand which month they're looking at.

## Steps to reproduce
1. Go to the home screen
2. Make sure you have some habits already added
3. Scroll horizontally on the days that you see on screen
4. As you scroll, you are not given information if the month changes making it impossible to know which month you're on

## System information
 - Phone: Samsung Galaxy S22
 - Phone Operating System: Android 13
 - App version: 2.0.3

## Screenshots
Attached
![Screenshot_20230728_110831_Habits](https://github.com/iSoron/uhabits/assets/24488614/3e663166-f78e-49b3-8731-6cc74a83962a)

kalina559:
Hi, I'll have a look into this :)

hiqua:
One explanation for this is that scrolling manually though many days using this view is somewhat not expected. A user looking through older records "should" click the habit and get the calendar views instead. Just saying that the not-so-great current design should not have that big of an impact.

exanrauzer:
The design is great, and the mentioned proposal is a natural enhancement and consistent with the user's intuitive flow.

torantal:
I'm just getting into the tool and find it very confusing that once I change the week (intentionally or unintentionally) there is no apparent way for me to see where I am or to get back to where today is. Why not show on the white space to the left of the days the month of the first day on the right?

moarsel:
This is an adjacent issue, but I also had no idea that I could scroll manually through the days to go back until I came and looked at the code! I agree that having the month in the header would be useful, and I think the behaviour should be enhanced so that a full swipe anywhere scrolls week by week. This is the common pattern used across mobile calendar apps.

PR #1780, opened 2023-07-29
Issue 1779: Showing month in HeaderView calendar

kalina559:
Solves #1779. Not sure about this one though, it could be useful when scrolling a long way back in the Header View calendar, but it basically shows the same thing for every day of the month. Below is the comparison of the current app (1), the code in this pull request (2) and a third option that would show the month only for first/last day of the month (3). Please let me know which one makes the most sense for you.

1.
MicrosoftTeams-image (22)

2.
MicrosoftTeams-image (23)

3.
MicrosoftTeams-image (24)


kalina559:
@iSoron @hiqua  I know a lot of tests started failing now, but before I fix them let me know if this change makes sense :)

hiqua:
The main drawback is that it takes more space, so you can visualize less habits at the same time. Maybe that's fine though? Wondering if we could have the month on the same row as the date instead. I don't have a strong opinion on this though.

iSoron:
Thanks for the PR, @kalina559. I'm not a huge fan of this layout as (i) it occupies more space; (ii) the month row will show the same text repeatedly, in the vast majority of situations. I agree that it would be nice to display the month sometimes, but I suggest doing what we currently do for the charts: on January 1, 2024, for example, we could show "Jan \\ 1" instead of "Monday \\ 1".

PR #1899, opened 2023-12-28
Unset measurable habits

leontodd:
I couldn't figure out a way to unset a measurable habit so I created this PR. If the save button is tapped in NumberDialog and the value is empty, the entry is saved as Entry.UNKNOWN. I've done it in line with how Entry.SKIP is saved.

Addresses #1722.


iSoron:
Thanks for the PR, @leontodd!

Issue #615, opened 2020-08-03
[feature] Color background for checkmark widgets that are ticked already 


AriApar:
First of all, amazing work with this app, been using it for some time now and am amazed at how integral it has become for my phone usage.

I had a small UX suggestion on the checkmark widgets. A habit can either be explicitly checked by a user, implicitly checked if the user already fulfilled the number of times they needed to check the habit in the last X days, or unchecked. When a feature is explicitly checked, the widget background also changes color. I was wondering if it'd make sense for an implicitly checked widget to also have some color (perhaps with some transparency)

![Screenshot_20200803-212530__01](https://user-images.githubusercontent.com/4266136/89215518-779ed300-d5d1-11ea-9d3b-d05cd5ef15d3.jpg)

In the screenshot above, the ~blog post~ (edit: this should have said "friends&family", thanks for the correction @wobbba) widget would be the one I'm talking about, and it'd be nice to look at my screen and immediately distinguish between the colorful vs grey boxes, to see which habits I should do today and which are already done (checkmark already provides this but is less in-your-face)

Thanks for considering, and again, amazing work with your app!

KristianTashkov:
To be honest I have the same consideration for the checkmarks in the app as well. On a dimmed screen you can't make the distinction(implicit check or unchecked) at a glance. It makes sense "color" to always mean success and playing with opacity to show the difference between explicit and implicit.

wobbba:
> I was wondering if it'd make sense for an implicitly checked widget to also have some color (perhaps with some transparency)
Came here to suggest this, so +1, that would be great!!

> In the screenshot above, the blog post widget would be the one I'm talking about
I'm confused, the blog post one isn't implicitly checked, it's the "friends & family" one, that one should have some colour IMO.

I also think it would be cool if the transparency and/or saturation and/or brightness of the colour would change according to how far away you are from breaking the chain. 

E.g. if it's a once a week habit and you just did it yesterday the colour is at its brightest/most saturated/most opaque and slowly fades day by day until it's at its most dark/unsaturated/transparent on the last implicitly checked day and then one day later it goes to completely colourless with an X, the way it is now.

[...]

PR #1915, opened 2024-01-05
Implicit Check widget color

wobbba:
As discussed in #615, I have implemented that the background color of the checkmark widget on implicit check (auto check) is now filled (colored background) like on manual check, but has an empty/stroked checkmark instead of a filled one. 

I have no experience with Kotlin/App-development but I tried to copy the patterns as best I could. It is a really small change and it seems to work, though I did only test it manually.

Basically, I only added the YES_AUTO to the same background color when-branch as YES_MANUAL and added an option to the RingView to make it draw the text as stroked with a stroke width dependent on the text size.

This is what it looks like now:
uhabits


iSoron:
Thanks for the PR, @wobbba. I have merged it in 45a82b3c2d2f13943167ccaa84f24852aed67072.

PR #1931, opened 2024-02-02
Fix: No notification, when Goaltype is set to "Maximum"

manish99verma:


iSoron:
@manish99verma Thanks for the PR.

PR #1960, opened 2024-03-19
Add option to change widget action

Peketr:
Hi, this pull request is of a feature implementation that has been bugging me since I installed this app.
Whenever I put a "checkmark widget" of a numerical habit to the home screen, I want to be able to increment it by a set amount on a tap instead of bringing up a full screen UI.

This pull request adds a new field to habits (increment) which tells the checkmark widget of that habit how much it should be incremented on tap. 
0.0 means the original behavior. 
I also added the UI for it in the Habit Edit UI.
image

For now, only an increment field decides what the Widget Action should be. This could be changed, but for now I didn't want to add 2 columns to the sql database.

Let me know what your thoughts are on this change/implementation.

### DIFF
--- a/uhabits-android/build.gradle.kts
+++ b/uhabits-android/build.gradle.kts
@@ -37,7 +37,7 @@ Updating gradle might fix this, so try again in the future to remove this and ru
 If this doesn't produce any warning, try to remove it.
  */
 kotlin {
-    jvmToolchain(11)
+    jvmToolchain(17)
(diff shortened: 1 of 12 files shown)


iSoron:
Thank you for the PR, @Peketr. This is an interesting idea, but I have decided to to merge it because the habit edit screen is already quite complex as it is, and this PR makes it even more complex.

PR #2012, opened 2024-06-25
New feature: allow users to automatically skip certain days of the week

dharux:
With this, the user can select days of the week to be skipped automatically every week. 

The use case I wrote this for is as follows: (closes #1980)
I have some tasks which I do only on working days and some tasks I do only on weekends/holidays. Therefore, I needed to automatically skip the weekend for some tasks and the working days for other tasks. To generalize it, any days of the week can be set for automatic skipping.

I added a new class, SkipDays, which is an attribute of the Habit class. It holds information on whether automatic skipping is turned on or off and which days of the week should be skipped. The skip days option is only available in the EditHabitAcitivity when it (isSkipEnabled) is enabled in preferences. It is also only enabled for daily and weekly habits. Auto-skipping days can be enabled in the edit habit activity. It uses the same weekday picker as the reminder to select the days to skip. 

Auto skipping days affects the following:
- HabitList: the automatically skipped days are marked as skipped for every week. If a user tries to add an entry for an auto-skipped day, the entry is blocked, and a toast message is displayed.
- TargetCard: for weekly numerical habits, the target is equally divided between the non-skipped days; so that the target is calculated properly for all intervals
- HistoryCard: the automatically skipped days are marked as skip for every week, starting from the day of the first manual entry, to prevent clutter. When the user tries to add an entry for an auto-skipped day, the entry is blocked, and a toast message is displayed.
- Streaks: the automatically skipped days are counted as part of a streak, so that a streak is not broken every week. However, any streaks with just skip entries are ignored. 

If the auto-skip feature is turned on for a habit with already existing entries, the entries are not modified in any way. If there already exists entries on the now auto-skip days, then those entries are simply ignored. 

The database version is bumped, and a new SQL file is added so that the update from the current version works.

I have also written some unit tests to test the streak and target card behavior with auto skip turned on.


iSoron:
Hi @dharux, thanks for the PR. I am unfortunately not planning to merge it for the following reasons:

First, I am not entirely convinced that this functionality is needed. You mentioned:

  > I have some tasks which I do only on working days and some tasks I do only on weekends/holidays. Therefore, I needed to automatically skip the weekend for some tasks and the working days for other tasks. 

The current supported way of achieving this is setting the habit to either 2 or 5 days per week. If you actually complete it, skips will be added automatically. If you don't, and it really bothers you to have some lapses on days in which you could not possibly complete the habit, you can add the skips manually.

Second, the PR proposes adding a new layer of complexity over our already-overly-complex non-daily scheduling of habits. The interaction between manual skips, automatic skips and the (newly-proposed) scheduled skips is way too complex, in my opinion, and I'm concerned that it will be an endless source of issues for the foreseeable future.

Finally, if you are proposing any changes to the user interface (such as creating new forms and screens), or implementing large changes to the codebase, such as this one, please prepare some mockups of your proposed changes and discuss it with us before writing any code, to avoid wasted effort. This is explained in our [contribution guidelines](https://github.com/iSoron/uhabits/blob/dev/docs/GUIDELINES.md#submiting-code).

PR #2022, opened 2024-07-17
Add repository now required for old kotlin-css-jvm versions

smowton:
Gradle Plugin Portal was previously making kotlin-css-jvm 1.0.0-pre.148-kotlin-1.4.30 available by virtue of mirroring JCenter. Now that Gradle have stopped doing that in preparation for JCenter to go away, we need to retrieve the old version from a different repository.

If this is bumped forwards to Kotlin 1.5.0+ it will become possible to get the package from Central instead.


iSoron:
Thanks!

PR #2028, opened 2024-07-28
Fix the misalignment of "About" page in RTL language

getgo-nobugs:
Hi,

I just found a bug related to the About page in Arabic.
Page in English

image

Page in Arabic

image

The root cause is that you did not put android:textAlignment="start" in the textview. This pull request fixed this issue. Now it looks good.

image


iSoron:
Thank you for the PR, @getgo-nobugs.

PR #2084, opened 2025-01-03
[Fix] Markdown linter problems

MacroController:
The Markdown linter was flagging some syntax problems. Fixed the MD024, MD033, MD034, MD036, MD040, MD045, and MD046 Markdown syntax problems.


iSoron:
Thanks!

PR #2089, opened 2025-01-12
Adding clear past entries feature (minor change)

Ankit-Jangid:
This can be added to clear the entries of a habit added by user instead of deleting a habit entirely.
Although, I know this is subjective and can have other opinions as well, other refinements are welcome. 

Reason: I once came across a scenario where I didn't want to delete the entire habit. but instead clear the entries so I can remove old data of that habit and wanted to start it over. 

Here's the feature:
(work is still in progress, it's not fully tested).

https://github.com/user-attachments/assets/38103954-00bc-4a25-ad4b-c07c457d3b6e

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/show/ShowHabitMenu.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/show/ShowHabitMenu.kt
@@ -56,6 +56,10 @@ class ShowHabitMenu(
                 presenter.onExportCSV()
                 return true
             }
+            R.id.action_clearEntries -> {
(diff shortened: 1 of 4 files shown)


iSoron:
Thank you for the PR, @Ankit-Jangid, but we have decided not to include this feature in the app. See https://github.com/iSoron/uhabits/issues/61#issuecomment-199942384 for more details.

Ankit-Jangid:
> Thank you for the PR, @Ankit-Jangid, but we have decided not to include this feature in the app. See [#61 (comment)](https://github.com/iSoron/uhabits/issues/61#issuecomment-199942384) for more details.

makes sense. thanks for explaining.  🙂

PR #2092, opened 2025-01-19
Option to inverse checkmark widget On/Off colours

JollyGood1:
Implemented the feature discussed in #2091.
Added option to choose to invert the checkmark widgets colours, making the colourful version a call to action.

Capture2
Capture

I'm not 100% sure about the what colour the UNKNOWN state should be because I don't really understand when it happens. I kept it the same as NO.

I've never contributed in github or ever coded for android, so feel free to tell me if I did something wrong.


iSoron:
Thank you for the PR, @JollyGood1; it looks well implemented. However, I don't think this feature should be added to the app. We are trying to keep the widgets consistent with the main interface. If anyone is interested in it, feel free to use this build.

PR #2093, opened 2025-01-21
Add otus directory with OtusIntroTest, OtusSettingTest, AboutScreen, …

yosemiu:
…IntroScreen, SettingScreen. Add androidTestImplementations for Kaspresso. Update testInstrumentationRunner for Kaspresso.


iSoron:
Thank you for the PR, @yosemiu, but is is unclear to me what it is trying to achieve.

Issue #2070, opened 2024-11-21
 "Hide Completed" mode with weekly/monthly measurable habits


miljenko20:
## Description
There is an issue with the "Hide Completed" mode when using measurable habits with **weekly and monthly** goals. Even when the cumulative weekly/monthly target is met or exceeded, the habit does not disappear from the list in "Hide Completed" mode. Instead, it seems to rely on whether a single day’s total meets or exceeds the entire weekly/monthly target, which is not expected behavior.
For example: If my weekly goal is 120 minutes of piano practice, the habit does not disappear even after accumulating more than 120 minutes over multiple days. It only disappears if I log 120 minutes or more on a single day.
This issue appears specific to weekly and monthly measurable habits and does not occur with daily measurable habits.

## Steps to reproduce
1. Create a measurable habit with a weekly goal (e.g., 120 minutes).
2. Log entries over several days to cumulatively reach or exceed the weekly goal, but do not log more than the goal value (120 minutes) on any single day.
3. Enable "Hide Completed" mode.
4. Observe that the habit remains visible, even though the cumulative weekly goal has been achieved.

## System information
Phone: Samsung Galaxy A32 4G
Phone Operating System: Android 13
App version: 2.2.1

## Screenshots
I have attached screenshots to help illustrate the issue. Please note that the app is in Croatian, but I believe the visuals should be clear enough to understand the problem. Let me know if you need any additional details or further clarification.

![WhatsApp Image 2024-11-21 at 22 58 26](https://github.com/user-attachments/assets/571c172e-bcf4-4b52-81bd-cd05f0df369c)
![WhatsApp Image 2024-11-21 at 22 58 26 (1)](https://github.com/user-attachments/assets/cb64993b-6dbf-4628-be7d-67aa4d8968b8)

Finally, I want to say how much I enjoy using the app. It is truly a fantastic tool—simple, intuitive, and highly customizable. I especially appreciate features like the ability to track progress visually and assign colors to habits. Thank you for your hard work and dedication to making this app so great!

Miljenko

Discostu36:
This also occured to me. It would be great if the app supported both types of goals.

GeorgeD88:
[...]

PR #2107, opened 2025-02-11
Hiding measurable habit if it's completed for the frequency period

kalina559:
This solves #2070. I added a check to see if the last frequency period satisfies the measurable habit's goal.

- DAILY - we're checking only today's entry
- WEEKLY - the sum of entries from the last 7 days including today
- MONTHLY - the sum of entries from the last 30 days including today

### DIFF
--- a/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/Habit.kt
+++ b/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/Habit.kt
@@ -72,7 +72,7 @@ data class Habit(
         val periodFirstDay = this.minus(frequency.denominator - 1)
 
         return computedEntries.getByInterval(periodFirstDay, this)
-            .filter { entry -> entry.value !=  Entry.UNKNOWN}
+            .filter { entry -> entry.value != Entry.UNKNOWN }
             .sumOf { entry -> entry.value }
     }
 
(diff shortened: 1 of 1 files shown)


iSoron:
Thanks for the PR, @kalina559, but unfortunately I think the solution needs to be more complex than this:

1. Note that we don't use rolling windows for numerical habits. See the target chart, for example; we group repetitions by week, month, quarter and year. For this reason, the first day of the week also matters for numerical habits. I think we should hide a habit on the main screen if the first row in the target chart indicates 100% completion (for weekly habits, the first row is "Week"; for monthly habits, it is "Month"). The good news is that we already have the code to compute this row, so we could just use it here, instead of re-implementing it, like you did in this PR. See `TargetCardPresenter.buildState`.
2. Skip days have an impact on the target. Specifically, if the goal is "reading 70 pages per week" and we skip one day, the target decreases to "63 pages" for that particular week. This is already handled in the existing code.
3. I'm a bit concerned about running the code in `TargetCardPresenter.buildState` for every habit on the main screen; it may be too slow. We need to try it out with large lists of habits, and refactor for performance if needed.
4. Ideally, we would not only hide the habit on the main screen, but also indicate through other means that the day is considered completed. I think this would include (i) marking the days as completed on the main screen, by making the number colorful, instead of black/grey; (ii) marking the day completed in the Calendar chart; (iii) marking the bars as completed in the History chart.

Issue #1857, opened 2023-10-27
0 / error while editing the count to zero, not going to passive rather stays in active. 


Divesh-Kumar-Chordia:
I entered in wrong date, so i removed it but it still considered it as active day. 
![Screenshot_2023-10-28-01-12-37-087_org isoron uhabits-edit](https://github.com/iSoron/uhabits/assets/88790423/b57c8978-3ddb-44be-9ac8-238bf089de23)

## Pre-submission checklist
- [ ] I am submitting a bug report, not a feature request.
- [ ] I have searched for similar issues, but did not find any matches.

## Description
A clear and concise description of what the problem is.

## Steps to reproduce
1. Go to ...
2. Click on ....
3. Scroll down to ....

## System information
 - Phone: [e.g. Google Pixel 4]
 - Phone Operating System: [e.g. Android 10]
 - App version: [e.g. 2.0.2]

caschual:
Same problem. Tried to find out whether there's a non-obvious solution, like removing the value (NULL) or replacing it with a "-" or something.
Nothing helped.

jozic:
Same issue, it shows as active on Calendar view and there's no way to reset the value to unset. It gets worse when you have a reminder for such a value where you want to count non zero values only. And once it pops up it seems like you have to enter something to satisfy it. And once you enter 0 it shows as done in the calendar.

KyleSCraig:
I believe this has been fixed by #2109 and is in version 2.3.0.

Issue #1815, opened 2023-08-21
Unable to set a question mark (missing data) or a dash (skip days) for measurable tasks. "Skip" defaults to "3"


xfarrow:
## Pre-submission checklist
- [X] I am submitting a bug report, not a feature request.
- [X] I have searched for similar issues, but did not find any matches.

## Description
Measurable tasks allow neither a dash (`-`) nor a question mark (`?`). Clicking on "skip" defaults to three (3). For example if I "skip" the question "how many miles have you walked today" it counts "3 miles".

## Steps to reproduce
1. Create a measurable task
2. Check the settings allowing dashes and question marks to be used
3. Notice their absence when inputing data
4. Click on "skip" and the data will be set to 3.

## System information
 - Phone Operating System: Android 13
 - App version: 2.1.2

## Screenshots

- Yes / No task:
![msg324741292-306268](https://github.com/iSoron/uhabits/assets/49845537/8f6bd258-673b-4fb5-8c60-3b1ca8674625)

- Measurable task
![msg324741292-306269](https://github.com/iSoron/uhabits/assets/49845537/e61aaa24-e72f-440b-bb1c-c56320d0ac11)

PR #2109, opened 2025-02-18
Added 'UNKNOWN' button for measurable habits

kalina559:
This solves #1857 and #1815 by enabling users to 'reset' an entry for a measurable habit if they added it by mistake. As this is not just a bug fix, but rather a new feature, I'll wait for an approval by @iSoron or @hiqua.

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/NumberDialog.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/NumberDialog.kt
@@ -47,7 +47,7 @@ class NumberDialog : AppCompatDialogFragment() {
             it.typeface = InterfaceUtils.getFontAwesome(requireContext())
         }
         if (!prefs.isSkipEnabled) view.skipBtnNumber.visibility = View.GONE
-        if (!prefs.areQuestionMarksEnabled) view.unknownBtnNumber.visibility = View.GONE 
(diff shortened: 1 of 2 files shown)


ZornHadNoChoice:
You guys are cooking recently, awesome work!

uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/NumberDialog.kt
```diff
@@ -36,16 +36,17 @@ class NumberDialog : AppCompatDialogFragment() {
         val appComponent = (requireActivity().application as HabitsApplication).component
         val prefs = appComponent.preferences
         view = CheckmarkPopupBinding.inflate(LayoutInflater.from(context))
-        arrayOf(view.yesBtn, view.skipBtn).forEach {
+        arrayOf(view.yesBtn).forEach {
```
oshhh, line 39: We don't have a yes button for number dialog right? Do we need this line at all?

hiqua:
Hey thanks, the code looks good to me, but I haven't tested it. About this, the fact that the tests pass with this change without any adaptation of the tests is a good indication that we are missing some tests... If you're willing to go the extra mile, adding a corresponding test would be nice.

iSoron:
Thanks for the PR, @kalina559. I have merged it with some minor changes.

PR #2121, opened 2025-03-21
Feature: Setting to allow for average of entries instead of sums

DasCapschen:
There are several discussions asking for a feature to allow them to show the average over weeks / months / years, instead of only sums. Personally, I also would like this feature, to track weight specifically. So... here's the MR :)

When creating a new habit, there is now a new setting "aggregation", where you can select either Sum (the default) or "Average" (new). This is only valid for Measurable Habits, not Yes/No habits. 
Values of SKIP and UNKNOWN are ignored in the calculation of the average. Only values the user entered are used to calculate the average. 

Example Screenshots

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/edit/EditHabitActivity.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/edit/EditHabitActivity.kt
@@ -44,6 +44,7 @@ import org.isoron.uhabits.activities.common.dialogs.WeekdayPickerDialog
 import org.isoron.uhabits.core.commands.CommandRunner
 import org.isoron.uhabits.core.commands.CreateHabitCommand
 import org.isoron.uhabits.core.commands.EditHabitCommand
+import org.isoron.uhabits.core.models.AggregationType
 import org.isoron.uhabits.core.models.Frequency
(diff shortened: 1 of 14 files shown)


iSoron:
Thank you for the PR, @DasCapschen; it's clear that a lot of work went into this. Unfortunately, I don't quite understand how this new aggregation option is supposed to work, and I suspect that it will also be very confusing to the users. Our current measurable habits were developed with the `sum` aggregation in mind, and the charts reflect this.

DasCapschen:
Well, it is apparent that you have no interest anymore to keep this project updated, thanks for making that crystal clear by closing the MR **BEFORE** having a discussion about how useful it may be.  

Yes, the app was designed with `sum` in mind, but do you really think I did not design this feature with `average` in mind? 
If you had taken one second to look at the screenshots, you would have seen, that with `average` selected, the charts now reflect that!   

Also, I assumed that you would be aware of the feature requests (discussions) on your own project, but maybe that's why you followed the trend of banishing unwanted things to Github discussions where less people will find them.
People there have already laid out how they expect this feature to work.  
https://github.com/iSoron/uhabits/discussions/1299

Well, it seems I will be maintaining my own fork of uhabits going forward.
Good day to you.

PR #2128, opened 2025-03-30
Added notes to CSV export

SimonCierniewski:
I add many notes to my Habit tracker. Later I review them during a weekly summary, but the CSV export was missing notes.

### DIFF
--- a/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/io/HabitsCSVExporter.kt
+++ b/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/io/HabitsCSVExporter.kt
@@ -122,9 +122,13 @@ class HabitsCSVExporter(
         val out = FileWriter(exportDirName + filename)
         generatedFilenames.add(filename)
         val dateFormat = DateFormats.getCSVDateFormat()
-        for ((timestamp, value) in entries.getKnown()) {
+        for ((timestamp, value, notes) in entries.getKnown()) {
             val date = dateFormat.format(timestamp.toJavaDate())
-            out.write(String.format(Locale.US, "%s,%d\n", date, value))
+            if (notes.isEmpty()) {
+                out.write(String.format(Locale.US, "%s,%d,\n", date, value))
+            } else {
+                out.write(String.format(Locale.US, "%s,%d,\"%s\"\n", date, value, notes.replace("\"", "”")))
+            }
         }
         out.close()
     }


uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/io/HabitsCSVExporter.kt
```diff
@@ -122,9 +122,13 @@ class HabitsCSVExporter(
         val out = FileWriter(exportDirName + filename)
         generatedFilenames.add(filename)
         val dateFormat = DateFormats.getCSVDateFormat()
-        for ((timestamp, value) in entries.getKnown()) {
+        for ((timestamp, value, notes) in entries.getKnown()) {
             val date = dateFormat.format(timestamp.toJavaDate())
-            out.write(String.format(Locale.US, "%s,%d\n", date, value))
+            if (notes.isEmpty()) {
+                out.write(String.format(Locale.US, "%s,%d,\n", date, value))
```
SimonCierniewski, line 128: I separated empty `notes` to reduce size of CSV file (avoid unnecessary `""`)

uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/io/HabitsCSVExporter.kt
```diff
@@ -122,9 +122,13 @@ class HabitsCSVExporter(
...
+            } else {
+                out.write(String.format(Locale.US, "%s,%d,\"%s\"\n", date, value, notes.replace("\"", "”")))
```
SimonCierniewski, line 130: `notes.replace("\"", "”")` is to correctly store notes with quotations `"`

druckdev: I'm not a contributor, just a user that is interested in this too, but:

Escaping any quotes instead of replacing them with a Unicode symbol might be a nicer solution.

[RFC-4180](https://www.ietf.org/rfc/rfc4180.txt) has to say about this:

> If double-quotes are used to enclose fields, then a double-quote appearing inside a field must be escaped by preceding it with another double quote.

This is also already done for the habit names in `Habits.csv`, by using `CSVWriter` (see [HabitList.kt](https://github.com/iSoron/uhabits/blob/dev/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/models/HabitList.kt#L181-L206)).

iSoron:
Implemented in 035b392ece8fa80c941996a016a9461e4fe1f9de in a more robust way, with tests.

PR #2129, opened 2025-03-31
Save notes when dialog closes (screen rotation, screen turned off, user accidentally clicked outside the dialog)

SimonCierniewski:
I extensively add notes to my tracked habits, and it's frustrating when my whole note disappears when I accidentally rotate the screen or my screen turns off.

I noticed that Dialogs are closed on purpose (`dismissCurrentDialog` in `onPause()`), because callbacks are cleared after rotation. Fixing it would involve architecture changes, so I added a quick fix instead to autosave notes when the dialog closes.

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/CheckmarkDialog.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/CheckmarkDialog.kt
@@ -38,10 +38,13 @@ import org.isoron.uhabits.utils.sres
 class CheckmarkDialog : AppCompatDialogFragment() {
     var onToggle: (Int, String) -> Unit = { _, _ -> }
 
+    private var wasSaved = false
+    private lateinit var view: CheckmarkPopupBinding
+
     override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
         val appComponent = (requireActivity().application as HabitsApplication).component
         val prefs = appComponent.preferences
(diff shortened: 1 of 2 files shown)


uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/CheckmarkDialog.kt
```diff
@@ -77,4 +79,17 @@ class CheckmarkDialog : AppCompatDialogFragment() {
 
         return dialog
     }
+
+    override fun onDestroyView() {
+        if (!wasSaved) {
```
SimonCierniewski, line 84: Don't save if user clicked one of values [✓, X, ? , -] to save the event.

uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/CheckmarkDialog.kt
```diff
@@ -77,4 +79,17 @@ class CheckmarkDialog : AppCompatDialogFragment() {
...
+            save(requireArguments().getInt("value"))
```
SimonCierniewski, line 85: If Dialog is closed by `dismissCurrentDialog()`, save note with previous value

iSoron:
Thank you for the PR, @SimonCierniewski, but we have decided not to save the notes unless the user actually confirms the dialog; this is what the Material Design guidelines recommend. A PR to actually restore the notes window after rotation would be welcome.

PR #2137, opened 2025-04-12
style: allow adding newlines in the "Name" field

powerjungle:
Since habit names are already shown with 2 lines in the main view, it makes sense to allow an user to add newlines to the names. This opens up the possibility for having better formatted text for the habit name.
As suggested by another user, having markdown support would also be a good addition at some point.

Closes: https://github.com/iSoron/uhabits/discussions/2101

Before:

Screenshot_20250412_192921_Habits

After:

Screenshot_20250412_193032_Habits


iSoron:
Thanks for the PR, @powerjungle. Merged!

PR #2138, opened 2025-04-13
New option that allows making completed/entered habits grey

powerjungle:
Works the same way as hiding habits, but only changes the color.

The reason this is useful, is because when a habit is treated as completed/entered, the cause might be just the frequency set for that habit.

So, if it's once per month and you've entered it only once, till a month has passed the habit will be hidden, but the habit might've been done again within that month and it would be useful to be able to see it to enter that data.

At some point it'll be good to change the darkest grey in the habit palet in "Edit habit", so that it doesn't conflict with the archived habit and "grey complete" colors.

It'll be good to call `finish()` on the settings activity, to avoid stacking activities every time the setting is changed.

Closes: https://github.com/iSoron/uhabits/discussions/2103

Option disabled:

Screenshot_20250413_192622_Habits

Option enabled:

Screenshot_20250413_192638_Habits

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsActivity.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/ListHabitsActivity.kt
@@ -75,8 +75,12 @@ class ListHabitsActivity : AppCompatActivity(), Preferences.Listener {
     private lateinit var menu: ListHabitsMenu
 
     override fun onQuestionMarksChanged() {
-        invalidateOptionsMenu()
-        menu.behavior.onPreferencesChanged()
+        if (prefs.greyCompleted) {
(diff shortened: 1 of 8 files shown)


nisc:
Sure, I like it. Should be optional, though.

powerjungle:
@nisc it is optional.

https://github.com/iSoron/uhabits/blob/d43fd52553da1738d41303a0e0af48d1d2b948a6/uhabits-core/src/jvmMain/java/org/isoron/uhabits/core/preferences/Preferences.kt#L104-L108

iSoron:
Thanks for the PR, @powerjungle, but I'm not planning on merging this functionality as it conflicts with our visualization of archived habits. Also, the "hide complete/entered" functionality we already have is supposed to cover your use case. If you do need to add a checkmark before the scheduled time, we made it easy to quickly disable "hide completed".

PR #2139, opened 2025-04-13
fix(ui): hacky fix for updating colors on pure black setting change

powerjungle:
This is to avoid restarting the app to see the resulting color change, when changing the pure black setting.

It'll be good to call `finish()` on the settings activity, to avoid stacking activities every time the setting is changed.


iSoron:
Thanks for the PR, @powerjungle. I agree that it would be nice to fix this, but the solution proposed in this PR has some issues. When you toggle the preference, the app unexpectedly opens the main activity, and when you close this activity using the back gesture, you are unexpectedly sent back to the settings page, instead of your phone's home screen.

The expected behavior when toggling the preference is for the changes to immediately apply to the (currently open) settings screen.

If you implement a new solution fixing this issue, feel free to submit a new PR.

PR #2143, opened 2025-04-19
Adds a way to change checkmark size in list view

lessershell:
Add checkmark scale value to preferences and allows a easy way to change the size of it.

check_mark_size_loop


iSoron:
Thank you for the PR, @lessershell. I'm not planning to merge this into the main app, as I think the alternative check mark sizes are not very aesthetically pleasing, but please feel free to use this in your own personal copy of the app.

PR #2158, opened 2025-06-07
Limit length of numerical habit units to 7

hiqua:


iSoron:
Thanks for the PR, @hiqua. I've implemented an alternative solution in e608c6ea623a9bc4483fe502952d00d9faf92f26 which actually measures the text and trims it as needed, instead of enforcing a constant character limit.

PR #2181, opened 2025-07-15
New Widget: Home Page Widget. A widget that looks like the homepage.

Skippyy22:
Hello! I've made a responsive widget that looks like the homepage but on a widget. (Called Habit List Widget)

The reason is because I wanted one, and saw a few others state that they wanted it as well. It follows the format of the other widgets as well and I tried my best not to change any other files and only to add functions when needed.

It follows the same logic as the homepage. When the widget is clicked you get taken to the homepage of the app.

The Stack widget used to use a multiple selector when selecting habits, so I revived that for this widget.

If a habit gets deleted, it simply does not show on the widget, HabitNotFoundException() is not thrown until all habits (in the widget) are null.

The bulk of the widget logic can be found here
uhabits-android/src/main/java/org/isoron/uhabits/activities/common/views/HabitListChart.kt

Closes: https://github.com/iSoron/uhabits/discussions/2117

Please reach out for any questions! This was a fun process.

Below is a quick video of the widgets responsiveness. And a picture for the widget's preview image.

https://github.com/user-attachments/assets/0dbd2f7d-a26f-43a2-bbc2-8a7a2b486694

widget_preview_habit_list


DaRealNim:
This is exactly what I'm missing from this app: a way to quickly check what habits remain to be done in the day from my home screen. Thank you for your work, hoping this gets approved!

One question: is the widget scrollable? If not, would it be easy to add this functionality? Sorry if that was mentioned somewhere and I missed it. It would be great to be able to check the entire habit list without the widget taking up the entire screen

Aravinth-Earth:
this is the exact feature i am looking for

just one query @Skippyy22 
1. does user needs to do selective picking of the habiits, or this widget can show all habits, with filter & sort options simailr to the home page feature set ?

iSoron:
Thanks for the PR, @Skippyy22. It's clear that a lot of work went into this. Unfortunately, I'm not planning to merge this or similar PRs for the following reasons:
1. Replicating the main app screen as a widget requires re-implementing a lot of existing functionality, as this PR shows, which is hard to get right. In this PR, for example, although the layout looks close, the spacing and sizing of elements is not quite right, which makes the widget looks unpolished. The widget is also missing a lot of the functionality we have in the main app, such as hiding entered habits.
2. This widget adds significant maintenance burden to the project. Any changes to the main app screen now also require updating the widget; we'd have to implement each functionality twice from now on, keeping in mind the limitations of Android widgets. This duplicated code is also now a potential location for bugs.
3. I think, for most users, the appeal of such a widget would be interacting with the app directly from the home screen, without opening Loop. This PR unfortunately does not solve that, since it still requires opening the app for any data entry.

Thanks again for your work, @Skippyy22, but I'm unfortunately not merging it.

jjhidalgar:
@iSoron 

And what about an alternative, less complex, version of this widget? Similar to the 1x1 widgets but with the status of every habit, and perhaps more customization.

Mainly, it would be nice to have an overview of all the habits and if "today" there is a need to complete one of them. I don't know, I feel like we are missing something that this PR attempted to solve. I understand the reason for closing, but still, some simplified summary of the habits seems to be needed on my home screen.

DaRealNim:
> Mainly, it would be nice to have an overview of all the habits and if "today" there is a need to complete one of them. I don't know, I feel like we are missing something that this PR attempted to solve. I understand the reason for closing, but still, some simplified summary of the habits seems to be needed on my home screen.

Agreed, a quick glance at "what is left for the day" is why I was excited about this PR

Skippyy22:
@iSoron all fair, I might try again and make it simpler and use more of the Home Screen functionality.
I was trying to follow the other widget functionality and that might have limited me, I may give this another go in the near future. Thanks for the feedback! 

1 I selfishly wanted more checks on the widget since it would be smaller than the home screen, but this causes a lot of unnecessary logic.

2. Although I'm calling a lot of functions for the checks and ring, that's fair, haha. 

3. I did not think about that, thanks. 

Also might make it scrollable too.

Thank you for the feedback will probably give it another go soon.

PR #2194, opened 2025-08-01
Added archive and unarchive to the habit statistics page

KyleSCraig:
Only one is visible at a time based on its current archive state. I did not run any of the tests, but I did check that it works.

* Only one option is available at a time and that option changes based on its archive state
* Archive does archive the habit, unarchive does unarchive the habit.

This implements #793 and #2191.

This is my first pull request ever.


iSoron:
Thanks for the PR, @KyleSCraig!

PR #2195, opened 2025-08-03
Show habit question when entering record

zenador:
For each habit, we are already allowed to write a question about it. We currently only see this question when looking at the overview page for the habit, but as a user, I would also like to see it on the pop-up when entering the daily record, so that I can add additional details to consider. If there is no question saved the pop-up will look the same as now. Not sure if others would also like this change though, so I could make this configurable if preferred.


iSoron:
Thanks for the PR, @zenador, but I'm not planning to merge it. Showing the questions here doesn't really seem necessary to me. I also think the design of the new dialog needs to be significantly improved.

PR #2198, opened 2025-08-07
Reverse order of noBtn and yesBtn

helskog:
When confirming a habit for the day the order of noBtn and yesBtn are in reverse. Adjusted the default view to conform with universal UI/UX standards and added a preference in the settings to flip the buttons back to the old view if people still want to use the old style.

Tested briefly in android studio myself to confirm functionality, hope this can get merged and pushed to your new update :) Thank you for the work you've done on this app!

https://github.com/user-attachments/assets/05fcba76-ed6e-4867-823d-357101fa291a


iSoron:
Thanks for the PR, @helskog, but I'm unfortunately not planning to merge it. Adding an option for this seems like overkill, and changing the default behavior now would be disruptive to existing users. I also don't see this screen as a regular dialog, with confirm/cancel buttons; it's more like a list of available states. In fact, if you enable skip days and questions marks, there would be four available options.

PR #2207, opened 2025-08-21
Automatic backups to public folder

MihanEntalpo:
## Added functionality to make automatic backups of SQLite database to the public folder.

**Why?**

I need a method to sync state of my habits with external scripts, so I wanted to make it save SQLite database copy to a public folder every 15 minutes, so that it can be picked up by Syncthing and sync with my laptop or under-the-desk home server. Also, I thought that this functionality would be useful for others, who don’t need such frequent backups, and want to set their own backup interval, so I've extended the setting for this case.

**Details of the PR:**

1) Added new settings to:
1.1) Enable/disable automatic backups
1.2) Select public folder for automatic backups, with permission request (If no public folder is selected, automatic backups cannot be enabled., and Toast appears when trying to do so)
1.3) Set the backup interval: every 15 min, 1 hour, 1 day, 3 days, 7 days, 14 days, 1 month
1.4) Set backup file name to use or not use datetime in its name

2) Added translations of all the strings to all the languages (for languages I don’t know, I used Google Translate, the strings are simple, so I think google translate should handle this well)

3) Added unit tests for new preferences. I don’t know how to test the automatic backup itself because I can’t force the app to assume that 15 minutes already passed in tests.

4) Modified already existing function DatabaseUtils::saveDatabaseCopy so that it can create backup files without datetime in its name, having previous behavior by default.

Attached screenshots
Screenshot_20250821_121101
Screenshot_20250821_121114


OdyAsh:
up. 

(love this btw, well done!❤️}

MihanEntalpo:
Update: found problem with MIUI, fixing...

PascalRomb:
Thank you!
I was working on my fork too but I had some troubles with SAF/uri and tests.

I tested your version on emulator with API 34 and saw that the autobackup scheduler works even if the application is closed (I will check on my device too, to be sure). Love it. I think I'll use yours, I like it more

did you implemented any kind of rotation policy? e.g. keeps only the last 5 elements in the folder, something like that

MihanEntalpo:
I've just encountered problem with backup on my Poco phone, and right now rewriting backup logic  using SAF, emulator doesn't have such limitation, so I didn't notice the problem.
Will update PR today.

No rotation policy is implemented, as I forgot about it :)
And my use-case is to backup file to the same name every time.

I think, I'll add option to set "how much backups should be kept", if SAF would allow this (not tried yet)

PascalRomb:
I did something [here](https://github.com/PascalRomb/uhabits/tree/feature/auto-export-backup)
Maybe you can reuse something. I did some tests too.

I just modified the AutoBackup logic already existing in order to use SAF. But I'm not sure it works on API 28/29.
The already existing logic is not using any scheduler at all, it just check the elapsed time since last backup.

The new logic will make a backup with date and another backup called "latest", to keep always the same name.

Note: I'm not an android/mobile developer, something could be wrong.

MihanEntalpo:
It seems, that you've done lots of work with GUI tests, I've skipped that part, because as I know where is no way to force Worker to think that 15 minutes already passed. I'll you your commits as a guide, and at least I can create GUI tests to configure automatic backups

MihanEntalpo:
Ok, now public auto-backups work fine on real POCO Phone, after I remade them via SAF.
I'll implement backups rotation to specified cound, and return with updated PR...

iSoron:
Hi @MihanEntalpo, thanks for the PR!

1. The app already has an auto backup feature, implemented in `org.isoron.uhabits.database.AutoBackup`, which also prunes old files. Could you just modify the existing feature to save the files at a user-provided SAF folder, instead of re-implementing that functionality?
3. This PR introduces too many options; we just need one, the backup location. Let's assume that backups are always on by default, that they are daily, and that they include the date; this is the current behavior. If the user does not select an external folder, it uses the current one, in the internal storage.

MihanEntalpo:
Hello, @iSoron , thanks for your reply.
But what should I do with my own need?
I need backup every 15 minutes not daily, to use it in automation, by synchronizing with remote computer.
The backup itself is a additional functionality. Also, it seems, that I'm not alone in such a use-case.

Let's assume, that I'll make like you've proposed, maybe additionally I'll add option to set backup frequency, that would be 1 day by default, and some other options, like 15min (it's smallest value for worker handler in android), 1 hour, 12 hours, and 1 day?

What do you think?
Otherwise this modification would be useless for me.

Also, I had an idea to make another change: make app store it's main database in a public folder, so, backup or sync could be made with any tools. Like: in options, we have field "Path to store main DB", if it is unset, store it in app's protected folder. If it set, main DB should be copied where, and used from the public folder.
This method has downsides like possible deletion or corruption of db, but user can do it on their own risk

iSoron:
> I need backup every 15 minutes not daily, to use it in automation, by synchronizing with remote computer.

Thanks again for the PR; but based on your last comment, it appears to me that the true purpose of this PR is to provide a partial implementation of a file-based sync function. I see several issues with this sync approach, so I'm not planning to pursue it. For example, saving the files every 15 minutes requires too much bandwidth, while being simultaneously too slow for real-time updates. I also don't think we should mix syncing with backups; they are completely different features.

If you're interested in modifying the PR to implement the functionality described in the title (i.e., simply allowing the user to choose an external backup location), I'd be glad to merge it. Otherwise, we can close it.

PascalRomb:
Hi, I've already done something similar [in this commits](https://github.com/PascalRomb/uhabits/commits/feature/auto-export-backup/)
But:

1. Some medium tests like IntentSchedulerTest not pass on my machine (apple silicon 1) when using emulated device arm64-v8a with API 28.
2. I did some changes to AutoBackupRun:
  - It's a daily backup but I'm doing double backup, one with date in the filename and one with "latest". 
  - Filename changed to "loop_habits_tracker_backup_(date|latest).db"
  - Default path is no path at all. I could not think a way to have the internal one as default. It seems to be inacessible when using SAF/DocumentUri. To keep both I need to have duplicated code: one that uses DocumentUri for selected path and one File path for internal one. :/
  - By default no auto backup at all (due to previous point)
  - I did not add the translation for UI strings :/

I also changed the test AndroidDatabaseTest because it failed from Android Studio but not from build.sh (but we can discuss later about it)

I tested on Nothing Phone 2a with Android 15.

If it's ok for you I can open a PR but please take a deep look before merging it because it's my first time with kotlin/Android studio.

MihanEntalpo:
> but based on your last comment, it appears to me that the true purpose of this PR is to provide a partial implementation of a file-based sync function

Ok, you right, this is some kind of sync and if there are some plans on implementing sync "the right way", my approach would be duplicated duct-tape semi-solution.

As I've already get this far, I will finish this PR so that it use built-in backup with public SAF folder (by the way, how would user disable public folder once they enabled it?)

Will think about other sync methods, while maintaining my own fork which will have an option to select backup frequency.

By the way, where we can discuss sync methods, so that I don't make large PRs which would not be merged? I saw low activity on proposed gitter chat. Should it be in discussions of GitHub?

MihanEntalpo:
I closing this PR because changes that are needed to make built-in backup use public folder is highly distinct from the features implemented here, so, I'll reopen new PR

Also, created discussion about sync methods
https://github.com/iSoron/uhabits/discussions/2208

MihanEntalpo:
new pull request is here: https://github.com/iSoron/uhabits/pull/2209

PR #2209, opened 2025-08-24
Automatic backup to public folder

MihanEntalpo:
## Added functionality to set a public folder for automatic backups of the app

**Details of the PR:**

1. Added a new setting to specify a public folder where automatic backups should be stored. If none is set, AutoBackup will use the default system app folder.

2. Added translations of all strings into all supported languages. For languages I don’t know, I used Google Translate. The strings are simple, so Google Translate should handle them well.

3. Added a test for the public backup feature.

Attached screenshot of the new setting: 
image

Created after this PR was declined: [https://github.com/iSoron/uhabits/pull/2207](https://github.com/iSoron/uhabits/pull/2207)


MihanEntalpo:
Hello, @iSoron! I've made it :) Existing AutoBackup functionality now can use public folder via SAF!

MihanEntalpo:
@iSoron I don't want to sound pushy, but just a reminder that the PR for backups to the public folder is ready for review :)

iSoron:
Thanks for the PR, @MihanEntalpo. I have tested it and it seems to work great; we can merge it as is. I will also continue to test it further in the next weeks. A couple of questions, if you happen to already know the answer:
1. Does this work with cloud-based folders, such as Google Drive or Dropbox?
2. Is it possible for the user to revoke access to the external folder, not from within Loop, but through the system settings or some other external method? What would happen in our code, then? Do we need to check that the authorization is still valid, and fall back to the internal folder?

MihanEntalpo:
Hello, iSoron, thanks for your response and merge.

> 1. Does this work with cloud-based folders, such as Google Drive or Dropbox?

No, right now it doesn't work, as it requires slightly more work to be done, and I wanted this PR to be as small as possible, so you can review it easily.

I think, I will make another PR which will use deeper SAF functionality, allowing to use third-side storage providers.

Right now easiest way to sync backup - is by using syncthing, or some kind of cloud storage that can sync user-selected folders, as it done in some of the NextCloud android clients.

> 2. Is it possible for the user to revoke access to the external folder, not from within Loop, but through the system settings or some other external method? What would happen in our code, then? Do we need to check that the authorization is still valid, and fall back to the internal folder?

File operations implemented in this PR doesn't require special access rights, and, I couldn't revoke right to access specified folder on my devices and in emulator. But where is a case when backups can stop working: if user remove the folder, or remove SD-card, where folder was specified. In that case backups would just post error messages to the log.

If you want, I can make a PR which would force app to "forget about public folder" if that folder had some number of errors during backup (as with SD card single error may not be a sign of failure, but be a result of temporary SD card ejection)

jjhidalgar:
Is there a release with this change included?

tstrijdhorst:
I am wondering this as well. I just forked the project and was poking around to create exactly this functionality. Then to my surprise it already seemed to exist?

Would love to use this!

ron-png:
@iSoron Do you have any information on when we can expect the next version of the App?
> Is there a release with this change included?

PR #2214, opened 2025-09-04
Fix #2210: Fix undoing the checkmark bug.

### LINKED ISSUE #2210: Undoing the checkmark is not possible when using a measurable habit
## Pre-submission checklist
- [ v ] I am submitting a bug report, not a feature request.
- [ v ] I have searched for similar issues, but did not find any matches.

## Description
When pressing skip or save on a measurable Habit (ie pages of a book read) it is not possible to remove it. There exists no (x mark) and returning the values to default (ie Zero pages) does not solve the problem. This happens though I didn't put in the first place any value or comment, but barely pressed save on a blank day
 
## Steps to reproduce
1. Go to a new measurable habit
2. Click on any day being in the calendar or the dashboard once to fill in the day Habit details (ie Number of pages read or Notes)
3. Press skip or save

## System information
 - Phone: Samsung Galaxy A32
 - Phone Operating System: Android 13
 - App version: 2.2.1

## Screenshots

Image
Image
Image
Image

KarthikeyanSathiyamoorthi:
Previously, when a user saved 0 for a measurable habit, the habit was incorrectly highlighted as completed, regardless of the actual target value. For example, a habit with a target of 10 would show as completed even when the user saved 0. This PR fixes the completion logic to only highlight habits when the saved value actually meets or exceeds the target value on both the Habits lists page and the Read page.

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/NumberButtonView.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/list/views/NumberButtonView.kt
@@ -173,8 +173,7 @@ class NumberButtonView(
 
         fun draw(canvas: Canvas) {
             val activeColor = when {
-                value < 0.0 -> lowContrast
-                (targetType == AT_LEAST) && ((value == 0.0) && (threshold != 0.0)) -> lowContrast
+                (value < 0.0) || ((targetType == AT_LEAST) && ((value == 0.0) && (threshold != 0.0))) -> lowContrast
                 (targetType == AT_LEAST) && (value >= threshold) -> color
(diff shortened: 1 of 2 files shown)


iSoron:
Thanks for the PR, @KarthikeyanSathiyamoorthi, but I unfortunately don't really understand the logic here. The changes only seem to apply to habits that have a target of zero, and tries to make a zero entry not look like it's complete. Why should that be the case? If the target is zero and the user entered zero, I believe the habit should be marked as completed. In other words, I think the current app behavior is correct.

What I question here is whether we should even allow zero targets at all, but I think there may be some useful applications and it's not causing huge issues, so I'm inclined to just leave it.

PR #2215, opened 2025-09-05
Feat: Add history average mode with include/exclude empty days option

Janssenhidal:
### Summary
This PR introduces new functionality for numerical habits, allowing users to
view averages inside the history card. Additionally, users can choose
whether empty days (zeros) are included or excluded from average calculations.

### Motivation
This feature addresses [#1299](https://github.com/iSoron/uhabits/discussions/1299),
providing more flexible and meaningful statistics for numerical habits.
Users can now:
- Compare average values across calendar weeks/months/years
- Decide whether missing days should dilute the average or not

### Screenshots
image

image

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/edit/EditHabitActivity.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/habits/edit/EditHabitActivity.kt
@@ -47,6 +47,7 @@ import org.isoron.uhabits.core.commands.EditHabitCommand
 import org.isoron.uhabits.core.models.Frequency
 import org.isoron.uhabits.core.models.Habit
 import org.isoron.uhabits.core.models.HabitType
+import org.isoron.uhabits.core.models.NumericalEmptyDaysMode
 import org.isoron.uhabits.core.models.NumericalHabitType
 import org.isoron.uhabits.core.models.NumericalHistoryType
 import org.isoron.uhabits.core.models.PaletteColor
@@ -86,6 +87,7 @@ class EditHabitActivity : AppCompatActivity() {
     var reminderMin = -1
     var reminderDays: WeekdayList = WeekdayList.EVERY_DAY
     var historyType = NumericalHistoryType.TOTAL
+    var emptyDaysMode = NumericalEmptyDaysMode.EXCLUDE_EMPTY
     var targetType = NumericalHabitType.AT_LEAST
 
     override fun onCreate(state: Bundle?) {
@@ -109,6 +111,7 @@ class EditHabitActivity : AppCompatActivity() {
             freqNum = habit.frequency.numerator
             freqDen = habit.frequency.denominator
             historyType = habit.historyType
+            emptyDaysMode = habit.emptyDaysMode
             targetType = habit.targetType
             habit.reminder?.let {
                 reminderHour = it.hour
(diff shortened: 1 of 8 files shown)


xdevs23 (changes requested):
Sorry for being away so long, was quite busy. Code looks good to me. I'd only suggest removing the two line changes that only affect spacing. This is to reduce the blast radius. Otherwise I'd say everything is exactly as expected.

Janssenhidal:
> Sorry for being away so long, was quite busy. Code looks good to me. I'd only suggest removing the two line changes that only affect spacing. This is to reduce the blast radius. Otherwise I'd say everything is exactly as expected.

Thanks for the feedback. Do you know of any best practices to remove the whitespace changes you pointed out in the easiest way?

xdevs23:
> > Sorry for being away so long, was quite busy. Code looks good to me. I'd only suggest removing the two line changes that only affect spacing. This is to reduce the blast radius. Otherwise I'd say everything is exactly as expected.
> 
> Thanks for the feedback. Do you know of any best practices to remove the whitespace changes you pointed out in the easiest way?

There are multiple ways to do it but here's a suggestion:

* Make sure you're on the branch that contains your changes and that you don't have any uncommitted changes.
* Determine the commit that needs changes first, that would be eeca98ae8982bc98dbbf468b487ccbcf374ddbc9
* Start an interactive rebase: `git rebase -i eeca98ae8982bc98dbbf468b487ccbcf374ddbc9^` (make sure you include the `^` which means "one commit prior")
* An editor appears with your changes. You'll want to turn two lines with "pick" into "edit", that would be eeca98ae8982bc98dbbf468b487ccbcf374ddbc9 (whitespace change in xml) and 2e2983fdb912a8824ed1836024cc01b4eda04e52 (removed line)
* Save and exit the editor, now the rebase begins and on the first edit it'll let you edit the commit. Proceed with the next step
* Edit the xml file and undo the whitespace change
* `git add -A` && `git commit --amend --no-edit` → now you added the change to the existing commit. Verify with `git show` that the whitespace change is indeed gone.
* You may perform as many edits as you'd like as mentioned above.
* `git rebase --continue` to move forward with the rebase
* You'll get the chance to edit the next commit, perform the editing step for the removed line and amend the commit as mentioned above.
* `git rebase --continue` → rebase moves forward
* Since you only changed two lines to "edit", all other commits are simply applied on top.
* Rebase complete, you may force-push your changes using `git push -f` which will be reflected in this PR

Rebasing means taking a **base** commit (in this case the one before the first change) and modifying the history that comes after it, effectively **re**writing the history that comes on top of the **base**.

Hope this helps.

Janssenhidal:
Thanks, I appreciate your thorough step-by-step advice. I **_think_** I implemented the changes you requested. To kindly confirm please.

xdevs23:
> Thanks, I appreciate your thorough step-by-step advice. I **_think_** I implemented the changes you requested. To kindly confirm please.

You're welcome. Excellent! You did it correctly. I appreciate it! Hopefully this gets merged soon.

iSoron:
Thanks for the PR, @Janssenhidal, but I am unfortunately not planning to merge it. Our current measurable habits are specifically designed for tasks that require a certain amount of effort daily, weekly, quarterly or yearly. Under this assumption, why would we need to show averages instead of sums, and why would we need to exclude empty days from the aggregation?

I think what this PR is implicitly trying to accomplish is to actually extend Loop to support other kinds of habits, such as weight tracking (the example in your screenshot), by making measurable habits more configurable. I unfortunately think this is the wrong approach. First, I'm not convinced we should expand the scope of the app to include weight tracking (suggested in this PR), or sleep tracking and finance tracking (as suggested in #1299); there are already multiple high-quality and very mature apps specifically for these applications, such as Google Fit and YNAB. Second, even if we do choose to expand the scope of the app to cover these applications, I think they should be implemented as new types of habits; adding more knobs to existing numerical habits just makes the app too confusing, especially for new users, in my opinion.

xdevs23:
@iSoron Thanks for letting us know about your opinion and the scope of the app.

See my response here: https://github.com/iSoron/uhabits/discussions/1299#discussioncomment-14400395

You suggested using apps like Google Fit and YNAB. Are you suggesting that we should use proprietary software instead of extending this free software? I am confused about that since this project is licensed under the GPL 3.0. Do you know about free/open source apps that can track sleep, finance and/or weight the way this PR suggests?

Notwithstanding the above, I would highly prefer using only one app instead of multiple. Should this not get merged, I'd rather fork the project and integrate this PR myself for my own use.

PR #2235, opened 2025-10-12
Fix crash in FrequencyChart when scrolling before 1970

Nixon2223:
Fix crash in FrequencyChart when scrolling before 1970, caused by a negative Unix timestamp being used in drawColumn()

-Added a defensive guard in drawColumn() to skip rendering if date.timeInMillis < 0.
-Ensures the app no longer crashes when users scroll far into the past.


iSoron:
Thanks for the PR, @Nixon2223, but I don't think this exception can be triggered through the actual UI.

PR #2242, opened 2025-10-27
Solve to Issue#2229

TangilHossain:
This is because date of unixtime starts from 1970. beyond that is neg, that is creating the crash.

Added a function in DateUtils.kt to calculate the months required to go to 1970 from today.
Then set the maxDataOffset to the number of months.
This limits the scrollableChart not to go beyond 1970.


TangilHossain:
@iSoron Please review the changes and tell me if any update requires.

iSoron:
Thanks for the PR, @TangilHossain. It looks good!

PR #2261, opened 2025-12-02
Auto-save notes when dismissing checkmark and number dialogs

fictiontoreality:
Previously, when a user entered notes for a habit entry, then dismissed the dialog without hitting a save button (either intentionally or, as often the case for me, accidentally), any changes to the notes would be lost.

This change seeks to avoid data loss by auto-saving notes when the dialog is closed, no matter the method. This handles several data loss scenarios:

- pressing the back button
- clicking outside the dialog box
- rotating the screen
- turning off the screen

(To be honest, I have done all of these at least once by accident, and some dozens of times.)

As part of this change, the NumberDialog now trims whitespace from notes when manually saving to be consistent with CheckmarkDialog.

### DIFF
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/CheckmarkDialog.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/dialogs/CheckmarkDialog.kt
@@ -20,6 +20,7 @@
 package org.isoron.uhabits.activities.common.dialogs
 
 import android.app.Dialog
+import android.content.DialogInterface
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View.GONE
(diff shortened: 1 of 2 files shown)


iSoron:
This was actually our initial implementation, but the material design guidelines at the time explicitly mentioned that touching outside of the dialog should "cancel the action and close the dialog", presumably not causing any data changes. It also said we should not add an explicit dismiss action. It appears that they no longer mention this in Material 3, so we could revisit auto save. 

Regardless, I think restoring the dialog after rotation makes sense and should be implemented. I would merge such a PR.

herrbrixter:
This PR is useful for Yes or No habits which doesn't have a Save button. Currently, there is no way to edit your notes without modifying your Yes or No choice. If you don't want to change your Yes or No choice, you have to pick your previous choice. (Ex. If your previous choice was Yes, you have to make sure to pick Yes again to make no change to your Yes or No choice.) Having to remember your previous choice just to save your notes is annoying.

This is not an issue for Measureable habits because it has a Save button.

fictiontoreality:
@iSoron I suspect the Material Design guidelines had in mind simple modal dialogs with only yes/no options and no text entry. In those cases, clicking outside the modal is essentially canceling the operation, which eliminates the need for a separate cancel button. However, more elaborate data entry - like composing notes or emails - typically uses full-screen dialogs that auto-save drafts to prevent data loss.

I can understand why Loop Habits uses a modal dialog: it allows quick data entry across multiple habits without excessive navigation. However, this approach doesn't optimize for the notes use case. The small text box makes it difficult to read and scroll through longer notes, and the modal's dismissibility creates multiple accidental data loss scenarios (rotation, back button, clicking outside, etc.).

For context on my use case: My notes often span multiple sentences or even paragraphs - similar in length to an email. Losing this content when I accidentally click outside the dialog is like having an email app discard my entire draft, forcing me to start over. Can be painful!

I see two potential paths forward:

* Long-term (P1): When the note field is focused, expand to a full-screen dialog (like email composers). When closed, return to the modal and save the note. This would balance quick habit entry with robust note editing.
* Short-term (P0): Accept this PR as-is. It prevents all data loss scenarios with minimal code changes while we consider a more comprehensive solution.

I'm happy to discuss either approach or explore alternatives you might have in mind. Thanks for considering this!

iSoron:
@fictiontoreality I have merged the PR. It looks good as is. Thanks again!

PR #2272, opened 2025-12-10
Linkify uris / links

Mathnerd314:
Per https://github.com/iSoron/uhabits/discussions/1634

There are really 2 options:
1. Add `android:autoLink="web|phone|email"` to XML (as done as part of first commit) - short, very self-contained.
2. Implement custom Linkify implementation (as done in final version of this commit tree)

The custom Linkify implementation offers more flexibility, in particular the possibility of supporting app-specific URLs like:
* `file:///sdcard/photo.jpg`
* `spotify:track:4cOdK2wGLETKBW3PvgPWqT`
* `vnd.youtube:dQw4w9WgXcQ`
* [Intent URIs](https://developer.chrome.com/docs/android/intents)

The issue with such flexible URIs is they may be spurious, hence use of queryIntentActivities. In practice this is restricted due to Android 11+ package visibility protection to prevent app sniffing, but QUERY_ALL_PACKAGES is set and also the app works fine without it, just is less powerful.

Linkify in standard Android source tree does not handle combination of standard autolinking + custom URI regex well, therefore the code is rewritten. Many package-private regexes are copied from android.util.Patterns but other than that the rewritten code is actually shorter and cleaner.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,7 +3,6 @@ org.gradle.daemon=true
 org.gradle.jvmargs=-Xms2048m -Xmx2048m
 android.useAndroidX=true
 android.enableJetifier=true
-android.defaults.buildfeatures.buildconfig=true
 android.nonTransitiveRClass=false
 android.nonFinalResIds=false
 org.gradle.configureondemand=true
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,5 +1,5 @@
 [versions]
-agp = "8.9.2"
+agp = "8.13.1"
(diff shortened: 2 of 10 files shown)


iSoron:
Thanks for the PR, @Mathnerd314, but I'm not planning on merging this functionality.

Mathnerd314:
Well, that is why it was a draft. I wanted comments on the approach, and some idea as to what would be merged.

Mathnerd314:
At this point it seems like you have completely rejected linkification as a feature with no explanation, even though it is a listed feature in the feature tracker.

PR #2313, opened 2026-02-14
Overlay multiple habit scores on single chart widget

jmangel:
## Summary

Implements the feature requested in #2264: when multiple habits are selected for the Score widget, overlay all score lines on the same chart using each habit's distinct color, instead of creating a swipeable stack of individual charts.

### UI Changes
- **Habit picker**: Score widget now shows a multi-select checkbox list with a Save button (other widget types are unchanged)
- **Chart rendering**: Multiple colored lines drawn on the same ScoreChart grid. Lines are drawn before markers to prevent XFERMODE_CLEAR artifacts between series.
- **Widget title**: Shows first habit name + count (e.g. "Exercise +2")

### Technical Changes
- `ScoreChart`: Add `ScoreSeries` data class and `setSeriesList()` method; backward-compatible `setScores()`/`setColor()` wrappers preserved
- `MultiScoreWidget`: New widget class taking `List<Habit>`
- `ScoreWidgetProvider`: Routes multi-habit case to `MultiScoreWidget` instead of `StackWidget`
- `HabitPickerDialog`: Add `allowMultiSelect()` hook; new `ScoreHabitPickerDialog` subclass
- New `widget_configure_activity_multi.xml` layout with checkbox list + Save button
- Updated `widget_score_info.xml` and `AndroidManifest.xml` for the new picker activity

### Notes
- This is AI-assisted (Claude Code)
- Looking for feedback on the UI approach before marking ready for review
- Existing single-habit Score widgets continue to work unchanged

## Test plan
- [ ] Add a Score widget → should show checkbox list with Save button
- [ ] Select 1 habit → renders single-line chart (backward compat)
- [ ] Select 3 habits → renders 3 colored lines on same chart
- [ ] Existing single-habit Score widgets continue to work after upgrade
- [ ] Other widget types (checkmark, history, etc.) unaffected

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/uhabits-android/src/main/AndroidManifest.xml
+++ b/uhabits-android/src/main/AndroidManifest.xml
@@ -113,6 +113,15 @@
         </activity>
 
         <activity
+            android:name=".widgets.activities.ScoreHabitPickerDialog"
+            android:exported="true"
+            android:theme="@style/Theme.AppCompat.Light.Dialog">
+            <intent-filter>
+                <action android:name="android.appwidget.action.APPWIDGET_CONFIGURE" />
+            </intent-filter>
+        </activity>
+
+        <activity
             android:name=".activities.about.AboutActivity"
             android:label="@string/about">
             <meta-data
--- a/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/views/ScoreChart.kt
+++ b/uhabits-android/src/main/java/org/isoron/uhabits/activities/common/views/ScoreChart.kt
@@ -45,6 +45,8 @@ import java.util.Random
 import kotlin.math.max
 import kotlin.math.min
 
+data class ScoreSeries(val scores: List<Score>, val color: Int)
+
 class ScoreChart : ScrollableChart {
     private var pGrid: Paint? = null
     private var em = 0f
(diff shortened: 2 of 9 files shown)


iSoron:
Thanks for the PR, @jmangel. Could you add a screenshot of the feature? My main concern with showing multiple lines in the same chart is clarity; how do you ensure that the user can tell which habit is which? I also think the lines might end up overlapping too much, making the visualization much less useful and confusing.

jmangel:
@iSoron Absolutely! Here's a screenshot
image

Great point! The implementation handles this by keeping backward compatibility — if only one habit is selected, it renders identically to the current widget. The multi-line mode is opt-in: the user chooses which habits to overlay via a multi-select checkbox dialog, so they control the density. Personally, I have 15 habits and currently need 15 separate Score widgets; grouping them into 3-5 widgets of 3-5 habits each would be a big improvement. Each line uses the habit's own color to distinguish them. That said, a legend or labels could be a good follow-up if clarity is an issue — happy to hear your thoughts!

jmangel:
Here's an idea for building a legend into the title, including a comparison showing what happens when there are too many habits to fit in the two-line-max title 
image 
image

jmangel:
Here is the before and after of my real-life use case:
image
image

PR #2335, opened 2026-03-28
Summary calendar

rsagie:
At the bottom part of the main screen, it displays calendar with all the activities that were done
Each with it own color.
Very similar to the calendar that is displayed on the second screen for each Habit.

Support settings toggle to display/hide the calendar, default is False (not display)


iSoron:
Thanks for the PR, @rsagie, but I'm not planning to merge this feature.

PR #2361, opened 2026-05-19
Fix timezone handling for local date display

Roh16632:
This PR replaces hardcoded GMT/UTC timezone handling with the system default timezone to ensure correct local date display for users in different regions.

Issue observed:
The app displayed an older date than the actual current local date.


iSoron:
@Roh16632 Thanks for the PR, but this is not an actual issue.

Issue #2344, opened 2026-04-13
Bug Report: Crash Due to Invalid String Input in FrequencyPickerDialog


JeffHuang04:
## Pre-submission checklist
- [x] I am submitting a bug report, not a feature request.
- [x] I have searched for similar issues, but did not find any matches.

## Description
The crash happens when the FrequencyPickerDialog is attempting to save input, likely expecting a numeric value (e.g., an integer), but instead, it gets a non-numeric string "tes".Since I am using an automated testing tool, the issue cannot be triggered manually.

## Steps to reproduce
1.Open the app and navigate to the FrequencyPickerDialog screen.
2.Try to save or enter a frequency value.
3.The app attempts to convert a string to an integer, but it fails when the input is "tes".
4.The app crashes with a NumberFormatException.

## System information
 - Phone: Android Emulator---Medium Phone
 - Phone Operating System: Android 9
 - App version: 2.3.1-Debug

## Screenshots
I don't have the screenshots.
## Debug info
```action: 4@click resource-id=org.isoron.uhabits:id/xTimesPerWeekRadioButton :android.widget.RadioButton@
action: 1@click resource-id=org.isoron.uhabits:id/everyDayRadioButton :android.widget.RadioButton@
action: 2@click resource-id=org.isoron.uhabits:id/everyXDaysRadioButton :android.widget.RadioButton@
action: 3@edit resource-id=org.isoron.uhabits:id/everyXDaysTextView :android.widget.EditText@3
action: 5@edit resource-id=org.isoron.uhabits:id/xTimesPerWeekTextView :android.widget.EditText@3
action: 6@click resource-id=org.isoron.uhabits:id/xTimesPerMonthRadioButton :android.widget.RadioButton@
action: 10@edit resource-id=org.isoron.uhabits:id/xTimesPerYDaysYTextView :android.widget.EditText@14
action: 7@edit resource-id=org.isoron.uhabits:id/xTimesPerMonthTextView :android.widget.EditText@10
action: 8@click resource-id=org.isoron.uhabits:id/xTimesPerYDaysRadioButton :android.widget.RadioButton@
action: 9@edit resource-id=org.isoron.uhabits:id/xTimesPerYDaysXTextView :android.widget.EditText@3
action: 11@click className=android.widget.Button,instance=0 :android.widget.Button@SAVE
--------- beginning of crash
04-13 05:29:06.972 21945 21945 E AndroidRuntime: FATAL EXCEPTION: main
[...]

PR #2379, opened 2026-07-18
Fix numeric validation in FrequencyPickerDialog

AlexIova:
I love the app and wanted to contribute. First time contributing to an open source project. 

I saw this issue https://github.com/iSoron/uhabits/issues/2344 and thought that was simple enough to tackle


iSoron:
Thanks for the PR, @AlexIova, but the bug itself is not valid.

